PR #7, opened 2023-07-19
Enforce ordered channel on `acknowledgement` and Migrate CoreSC tests to Forge

alfredo-stonk:
Main changes:
- enforce ordered channel checks in `CoreSC.acknowledgement` method; `recvPacket` already has checks in place
- migrate Hardhat tests to `forge`
- updated GH workflow so `forge test` is run 

NOTE: old Hardhat tests are removed to avoid dup


contracts/Dispatcher.sol
```diff
@@ -548,6 +519,16 @@ contract Dispatcher is IbcDispatcher, Ownable {
         bool hasCommitment = sendPacketCommitment[address(receiver)][packet.src.channelId][packet.sequence];
         require(hasCommitment, 'Packet commitment not found');
 
+        // enforce ack'ed packet sequences always increment by 1 for ordered channels
```
qinlz2, line 522: The `ack` in IBC term is mixing the concept of a transport layer ack and application layer ack.   So it is dual role. 

- For transport layer, `ack` (such as TCP acks), should be received sequentially.   

- For application layer, `ack` doesn't need to be sequential. 
   -  Some messages doesn't have application ack at all. 
   -  IBC ack can work in `async` mode natively. 
       - These results the ack won't be **always increase by 1**

qinlz2 (changes requested):

contracts/Dispatcher.sol
```diff
@@ -657,6 +635,7 @@ contract Dispatcher is IbcDispatcher, Ownable {
         require(!hasAckPacketCommitment, 'Ack packet commitment already exists');
         ackPacketCommitment[address(receiver)][packet.dest.channelId][packet.sequence] = true;
 
+        emit RecvPacket(address(receiver), packet.dest.channelId, packet.sequence);
```
qinlz2, line 638: This is the duplicate logic for line 611, so we will see duplicate `RecvPacket` events as receipts.

contracts/Ibc.sol
```diff
@@ -45,3 +58,10 @@ struct PacketFee {
     uint256 ackFee;
     uint256 timeoutFee;
 }
+
+struct Proof {
+    // block height at which the proof is valid for a membership or non-membership at the given keyPath
+    uint64 proofHeight;
```
qinlz2, line 64: This is ok for now.   (As we use this single integer `height` for polymer height)

- We have various places use integer height. 
- This means, if we do `polymer` upgrade, it will be hard to 
   increase `polymer revision_number`  (hard fork). 

We will see if we need that revision_number bump when we do first staging chain upgrade.   

If we do need revision_number, then we need to improve multiple components to support that (revision, height) pair.

forge-cache/solidity-files-cache.json
```diff
@@ -0,0 +1,1692 @@
+{
```
qinlz2, line 1: Do we need to put this cache.json into repo? 
Or maybe can `ignore` it?

contracts/Dispatcher.sol
alfredo-stonk: Ordered Acks are enforced for ordered channel. Dapps can choose unordered channel if ordering is not critical to them. What we did here is at the transport layer handled by CoreSC, same as IBC module on IBC chains https://github.com/cosmos/ibc-go/blob/6bfa0b3ea997c1a6c1f99a055935ee170f0ff129/modules/core/04-channel/keeper/packet.go#L443

contracts/Ibc.sol
alfredo-stonk: To be future proof, I changed this to `Height { reversion_number, revision_height } `

contracts/Dispatcher.sol
qinlz2: Great clarification.   Yes, I re-checked the spec, it does have similar `functional semantics` requirement for ack to be always generated for ordered channel.

PR #12, opened 2023-07-27
encoding test and IBC height conformance 

alfredo-stonk:


alfredo-stonk:
Current dependencies on/for this PR:
* main
  * **PR #12** <a href="https://app.graphite.dev/github/pr/open-ibc/vibc-core-smart-contracts/12" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/open-ibc/vibc-core-smart-contracts/12?utm_source=stack-comment).

contracts/payload/proto/gogo.proto
```diff
@@ -0,0 +1,145 @@
...
+// OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+// LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+syntax = "proto3";
+package gogoproto;
```
dshiell, line 30: can this file be imported? It looks like in polymer protos we use `import "gogoproto/gogo.proto";`?

.gitmodules
```diff
@@ -4,3 +4,6 @@
 [submodule "lib/openzeppelin-contracts"]
 	path = lib/openzeppelin-contracts
 	url = https://github.com/openzeppelin/openzeppelin-contracts
+[submodule "lib/protobuf3-solidity-lib"]
```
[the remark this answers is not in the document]
dshiell: or possibly npm install requirement?

contracts/Dispatcher.sol
```diff
@@ -79,13 +79,7 @@ contract Dispatcher is IbcDispatcher, Ownable {
...
     );
 
-    event WriteTimeoutPacket(
-        address indexed writerPortAddress,
-        bytes32 indexed writerChannelId,
-        uint64 sequence,
-        uint64 pktTimeoutHeight,
-        uint64 pktTimeoutTimestamp
-    );
+    event WriteTimeoutPacket(address indexed writerPortAddress, bytes32 indexed writerChannelId, uint64 sequence);
```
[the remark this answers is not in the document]
qinlz2: They are used in the correct flow.   So please don't remove them. 

- pktTimeoutHeight / Timestamp are stored on polymer chain in a specific block.   `eth2-relayer` is able to scan those if needed, but it will be a harder work, and need to look very deep (because no clue about where the block is, but this is a timeout, so the block might be far lower than current height)

- So the current flow is letting ETH side provide this data directly.

contracts/Dispatcher.sol
```diff
@@ -279,7 +273,7 @@ contract Dispatcher is IbcDispatcher, Ownable {
         if (
             counterparty.channelId != 0 ||
             bytes(counterparty.version).length != 0 ||
-            proof.proofHeight != 0 ||
+            proof.proofHeight.revision_number != 0 ||
```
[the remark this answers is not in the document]
qinlz2: Is revision_number starting from 1?   I thought it is starting from 0.    If so, do we need this check?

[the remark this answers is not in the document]
qinlz2: But because `proof` structure got changed, then relayer are related component need updates.

contracts/Dispatcher.sol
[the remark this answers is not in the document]
alfredo-stonk: These fields are part of the original packet. So relayers don't need them to id packets

contracts/Dispatcher.sol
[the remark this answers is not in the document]
alfredo-stonk: This's meant to be a zero-value check, so both are required. But I refactored the proof verification condition, so it's more clear ChanOpenInit and ChanOpenTry are handled separately

.gitmodules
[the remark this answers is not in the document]
alfredo-stonk: .gitmodules is actually managed by `forge`. `forge test/compile` will auto download deps. I added instructions in Readme in last PR :) 
Contract tests with hardhat were removed. We can get rid of npm project completely later

contracts/Dispatcher.sol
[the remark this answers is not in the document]
alfredo-stonk: Just saw @qinlz2's comments. If it's still used, I can bring them back. 
IMO we should delete them later

contracts/Dispatcher.sol
```diff
@@ -83,8 +83,8 @@ contract Dispatcher is IbcDispatcher, Ownable {
         address indexed writerPortAddress,
         bytes32 indexed writerChannelId,
         uint64 sequence,
-        uint64 pktTimeoutHeight,
-        uint64 pktTimeoutTimestamp
+        Height timeoutHeight,
```
qinlz2, line 86: We can use `Height` type, or we can use `uint64`. 

There is no much difference unless an EVM chain support revision as well.  (I doubt it).

It's fine to keep `Height` here.

contracts/Dispatcher.sol
```diff
@@ -276,12 +276,16 @@ contract Dispatcher is IbcDispatcher, Ownable {
...
-            proof.proof.length != 0
-        ) {
+        bool isChanOpenTry = false;
+        if (counterparty.channelId == bytes32(0) && bytes(counterparty.version).length == 0) {
+            // ChanOpenInit with unknow conterparty
+        } else if (counterparty.channelId != bytes32(0) && bytes(counterparty.version).length != 0) {
+            // this is the ChanOpenTry; counterparty must not be zero-value
+            isChanOpenTry = true;
+        } else {
+            revert('Invalid counterparty');
```
qinlz2, line 286: Good to see `revert` is used here, so I can learn what that is, and what's the difference between `require` and `revert`
When search online (correct me if I read them wrong).

- `require`: take caller's gas, meaning caller makes a mistake therefore they experience gas loss.
- `revert`: do not take caller's gas, meaning this is a situation typically Solidity contract has buggy logic.   Since this is SC's bug, do not take caller's gas.

==================================
If above understanding is correct (please consult more experienced SOL developer in our team), 
- I think `revert` shouldn't be used here, since this is a caller's error.  (We should take their gas).
- We probably wish to re-examine our SOLs for `require` and `revert` usage, to make sure they are correctly used.

contracts/Ibc.sol
```diff
@@ -10,9 +12,25 @@ struct IbcPacket {
...
+/// that can be compared against another Height for the purposes of updating and
+/// freezing clients
+///
+/// Normally the RevisionHeight is incremented at each height while keeping
+/// RevisionNumber the same. However some consensus algorithms may choose to
+/// reset the height in certain conditions e.g. hard forks, state-machine
+/// breaking changes In these cases, the RevisionNumber is incremented so that
+/// height continues to be monitonically increasing even as the RevisionHeight
+/// gets reset
+struct Height {
```
qinlz2, line 31: Just a note here, not a comment:

`Height` here is a general way to represent block chain height.
- When `Height` represents a cosmos height (e.g. LC's field for Polymer height), `revision_number` has a meaning.
- When `Height` represents a EVM height, `revision_number` is undefined.

contracts/payload/hash.sol
qinlz2, line 1: These files under `payload` folder seems for this PR. 

Should we included them for this PR?

qinlz2:
A general comment:
The changes for `height` will impact both vibc relayers.   
If we merge, we need to have compatible code ready from relayer side.

contracts/Dispatcher.sol
alfredo-stonk: Let's do that in another PR? I added an issue #13. Custom errors seem better than error strings

contracts/Dispatcher.sol
alfredo-stonk: True.
OTOH, when `WriteTimeoutPacket` is emitted, the running chain is B.  The timeout params here are from the original packet on chain A. So we include them as-is for completeness. Whether the `revision_number` makes sense or should be 0 are determined by IBC clients on Polymer.

contracts/payload/hash.sol
alfredo-stonk: payload folder is for codec testing. Do you want me to split this PR into 2?

alfredo-stonk:
> A general comment: The changes for `height` will impact both vibc relayers. If we merge, we need to have compatible code ready from relayer side.

Right. I'm also working on ibc-29, which would cause changes to vibc relayers. We can defer tagging any CoreSC version until ibc-29 work is done?

dshiell (approved):
lgtm, we can refine as needed

contracts/Dispatcher.sol
dshiell: If I understand @qinlz2's concern we could simply replace 'revert' with 'require' to prevent spamming this call with invalid counterparty data?

dshiell: nm, this explains it pretty well: https://medium.com/blockchannel/the-use-of-revert-assert-and-require-in-solidity-and-the-new-revert-opcode-in-the-evm-1a3a7990e06e

revert == require but for more complex conditionals like we have here. It allows refunding unspent gas which is good.

assert is for hard invariants - situations which should never occur

dshiell (approved):

nicopernas (approved):
Looks good. We are going to have to make changes to both relayers before the e2e is ready.

PR #16, opened 2023-08-01
Ds/migrate to npm registry

dshiell:
Remove registry override and bump version to 0.0.10

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@open-ibc/vibc-core-smart-contracts",
-  "version": "0.0.9",
+  "version": "0.0.10",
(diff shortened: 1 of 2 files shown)


nicopernas:
Do we also need to update the publish workflow so the new packages are published to npmjs and not gh?

dshiell:
the default is npmjs

```derek@Dereks-MacBook-Pro-2 vibc-core-smart-contracts % npm config ls -l | grep registry
metrics-registry = "https://registry.npmjs.org/" 
omit-lockfile-registry-resolved = false 
**registry = "https://registry.npmjs.org/"** 
replace-registry-host = "npmjs" ```

Issue #11, opened 2023-07-26
[Proposal] Optimistic ZK connections


zhengyangfeng00:
## Background

The objective is to decrease the gas cost of verifying ibc related messages. For additional background, refer to the design document [ADR 10 Optimistic ZK Connections](https://www.notion.so/cb54af27d51345fa8ede6c5020566b67?pvs=21). This issue aims to discuss the implementation level design, focusing on the modifications that need to be made on the vibc core SC side.

### Terms

- **Non-Optimistic consensus state**: This is a subset of the block header that includes the public keys of all active validators, as well as their voting power and other relevant information.
- **Optimistic consensus state**: This is the Merkle root of the Polymer chain's state. It can be used to perform inclusion proofs for any key-value pairs stored on the Polymer chain. The Op consensus state is part of the Non-Op consensus state.
- **Client update**: This refers to updating the Polymer chain's consensus state. There are two types of client updates:
    - Updating the non-OP consensus state: The non-OP consensus state can be verified on-chain and is considered trusted immediately after passing the on-chain verification.
    - Updating the OP consensus state: The OP consensus state is buffered and not verified on-chain. Instead, it relies on watchers to perform off-chain fraud detection. Once the fraud proof window has passed, the OP consensus state is considered trusted and can be used to perform inclusion proof for data packets.
- **Packet verification**: This refers to using Merkle proof to verify the existence of an IBC data packet in relation to a consensus state at a specific height, among other things.
- **Watcher**: This component is responsible for checking for fraudulent client updates or fraudulent data packets.
- **Updater**: This component is an off-chain process that performs client updates periodically. Its purpose is to ensure that the non-op consensus state of the vIBC core SC is not more than one bounding period older than the latest block on the Polymer chain.

There are two approaches for implementing the optimisitc packet routing, with different pros and cons.

### High level

There are 4 roles in the client state update process:

- the relayer
- the watcher
- the vibc core SC
- the polymer chain

The interactions between these components can be illustrated with the diagram below:

<img width="1124" alt="image" src="https://github.com/open-ibc/vibc-core-smart-contracts/assets/1428290/7e33207c-bd09-4d1b-86d4-62de203c150b">

The sequence diagram above describes two different cases:

**Case 1: Relaying a data packet when fraud is not detected**

1. The relayer sends an Op consensusState update to the smart contract. The vIBC core smart contract saves the Op consensusState in its internal state and waits for the fraud proof window for the state to pass.
[...]

PR #17, opened 2023-08-01
feat: support optimistic packet relaying

zhengyangfeng00:
this implements the design in #11 

changes:
- split the consensus state into consensus state and optimistic consensus state
  - consensus state is used for verifying fraud proof, while optimistic consensus state is used for membership proof for incoming IBC packages
  - consensus state includes more fields than the optimistic consensus state which is just a few hashes
  - at the current impelementation the consensus is simplified and is the same as the optimistic consensus state
- combine proof height check and the membership verification into a single function; the function is called in methods that receive ibc messages 
- add proof window check before using the op consenus state for membership verification
- added basic tests; planning to add more in follow up PRs


zhengyangfeng00:
offline discussed with @alfredo-stonk , here is the summary:

Todos

in this pr:
- refactor the union to use different methods
- naming change, consensusState remains the same, clientState -> OptimisticConsensusState
- add a state to allow locking / unlock this contract

in the future:
- refactor the code to have a CLientState contract

alfredo-stonk (approved):

PR #25, opened 2023-08-25
feat: try to promote pending op consensus state from more methods

zhengyangfeng00:
this fixes the following case:
- vibc relayer needs to check the latest consensus state to determine if a client state update is necessary before relaying a packet. it does so by calling the `getTrustedOptimisticConsensusStateHeightWithPromotion` method on the dispatcher contract.
- the dispatcher contract would try to promote the pending op consensus state before returning the current trusted one.


dshiell (approved):
lgtm, only small thing to consider is the getter func `getPendingOptimisticConsensusStateHeightWithPromotion` possibly mutating state, but could be ok

PR #26, opened 2023-08-29
feat: support storing op consensus states of multiple heights

zhengyangfeng00:
## high level
- encapsute the op consensus state update logic in a separate class
- support querying the op consensus state at a specific height

related to #11 

## gas cost estimation
gas cost estimation for `addOpConsensusState`

- case: adding a new appHash, 50k
- case: adding the same appHash again, 1.6k

```
| test/OpConsensusStateManager.t.sol:OptimisticConsensusStateManagerCaller contract |                 |       |        |       |         |
|-----------------------------------------------------------------------------------|-----------------|-------|--------|-------|---------|
| Deployment Cost                                                                   | Deployment Size |       |        |       |         |
| 156501                                                                            | 719             |       |        |       |         |
| Function Name                                                                     | min             | avg   | median | max   | # calls |
| addOpConsensusState                                                               | 50547           | 50547 | 50547  | 50547 | 1       |
```

```
| test/OpConsensusStateManager.t.sol:OptimisticConsensusStateManagerCaller contract |                 |       |        |       |         |
|-----------------------------------------------------------------------------------|-----------------|-------|--------|-------|---------|
| Deployment Cost                                                                   | Deployment Size |       |        |       |         |
| 156501                                                                            | 719             |       |        |       |         |
| Function Name                                                                     | min             | avg   | median | max   | # calls |
| addOpConsensusState                                                               | 1608            | 26077 | 26077  | 50547 | 2       |
| getState                                                                          | 1799            | 1799  | 1799   | 1799  | 1       |
```

gas cost estimation for `getState`

- 1.80k in all cases

### DIFF
--- a/contracts/OpConsensusStateManager.sol
+++ b/contracts/OpConsensusStateManager.sol
@@ -1,0 +1,53 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.0;
+
+import './IbcVerifier.sol';
+
+
+// OptimisticConsensusStateManager manages the appHash at different
+// heights and track the fraud proof end time for them.
+library OptimisticConsensusStateManager {
+    struct DataStore {
+        uint32 fraudProofWindowSeconds;
+
+        // consensusStates maps from the height to the appHash.
+        mapping(uint256 => uint256) consensusStates;
+
+        // fraudProofEndtime maps from the appHash to the fraud proof end time.
+        mapping(uint256 => uint256) fraudProofEndtime;
+    }
+
+    // addOpConsensusState adds an appHash to internal store and
+    // returns the fraud proof end time, and a bool flag indicating if
+    // the fraud proof window has passed according to the block's
+    // timestamp.
+    function addOpConsensusState(DataStore storage self, uint256 height, uint256 appHash)
+        external returns (uint256 fraudProofEndTime, bool ended) {
+        uint256 hash = self.consensusStates[height];
+        if (hash == 0) {
+            // a new appHash
+            self.consensusStates[height] = appHash;
+            uint256 endTime = block.timestamp + self.fraudProofWindowSeconds;
+            self.consensusStates[height] = appHash;
+            self.fraudProofEndtime[appHash] = endTime;
+            return (endTime, false);
+        }
+
+        if (hash == appHash) {
+            uint256 endTime = self.fraudProofEndtime[hash];
+            return (endTime, block.timestamp > endTime);
+        }
+
+        revert('cannot update a pending optimistic consensus state with a different appHash, please submit fraud proof instead');
+    }
+
+    // getState returns the appHash at the given height, and the fraud
+    // proof end time.
+    //
+    // 0 is returned if there isn't an appHash with the given height.
+    function getState(DataStore storage self, uint256 height) public
+        returns (uint256 appHash, uint256 fraudProofEndTime, bool ended) {
+        uint256 hash = self.consensusStates[height];
+        return (hash, self.fraudProofEndtime[hash], hash != 0 &&  block.timestamp > self.fraudProofEndtime[hash]);
+    }
+}
--- a/contracts/Dispatcher.sol
+++ b/contracts/Dispatcher.sol
@@ -9,6 +9,7 @@ import './IbcDispatcher.sol';
 import './IbcReceiver.sol';
 import './IbcVerifier.sol';
 import {Escrow} from './Escrow.sol';
+import './OpConsensusStateManager.sol';
 
 // InitClientMsg is used to create a new Polymer client on an EVM chain
 // TODO: replace bytes with explictly typed fields for gas cost saving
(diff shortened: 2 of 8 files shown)


contracts/OpConsensusStateManager.sol
```diff
@@ -0,0 +1,127 @@
...
+    // adding a duplicate OptimisticConsensusState is an idempotent
+    // operation. it doesn't change the fraud proof end time or change
+    // the state of an already trusted OptimisticConsensusState.
+    //
+    // it is an error to update a pending OptimisticConsensusState
+    // with different values. such operation must be done through the
+    // sumitting fraud proof API.
+    //
+    // the status and the fraud proof end time are returned.
+    function addOpConsensusState(DataStore storage self, OptimisticConsensusState calldata state)
```
notbdu, line 41: Can you profile the gas costs of these fns?

Would try to simplify logic where we can. On-chain costs can be quite high.

contracts/OpConsensusStateManager.sol
```diff
@@ -0,0 +1,127 @@
...
+        s.status = OpConsensusStateStatus.PENDING;
+        // n.b, don't trust the timestamp sent by the caller.
+        s.state.time = block.timestamp;
+        self.heights.push(state.height);
+        self.size++;
+
+        // delete the op consensus state with the smallest
+        // height(i.e., oldest) from internal store.
+        uint256 smallestHeight = self.heights[0];
+        if (self.size > self.capacity) {
```
notbdu, line 75: Shift operations are incredibly expensive in the EVM. Every value that gets updated costs 20k gas.

A ZK update is ~330k gas, the more gas this optimistic update costs the less it even makes sense to invest the engineering effort into doing this. We want to optimize for minimal gas overhead.

contracts/OpConsensusStateManager.sol
```diff
@@ -0,0 +1,127 @@
...
+// states. It keeps track a configurable number of optimistic
+// consensus states. IBC module uses this class to access an
+// optimistic consensus state at a given height to perform membership
+// proof.
+library OptimisticConsensusStateManager {
+    enum OpConsensusStateStatus { NOT_APPLIED, PENDING, TRUSTED }
+
+    struct ManagedOpConsensusState {
+        OptimisticConsensusState state;
+        OpConsensusStateStatus status;
```
notbdu, line 17: Can prob get rid of status field completely. Time inserted is enough info on whether we can use a state for proofs.

notbdu:
High level comments are:
- Cannot write solidity like writing normal code (EVM ops are incredibly expensive)
- For example, adding a new storage slot (single value, complex data structures cost per field) costs 20k gas per value.
  - This is ~$1 per inserted value at current eth price
- Do not need to record/do everything on chain (some data can be emitted as events and indexed off chain to inform optimal relayer behaviour)
  - Enforce critical paths on chain only
  - Some data may not be needed, we only care about appHash for proving packet proofs. May be cheaper overall to only store a hash of the entire consensus state obj and pass in data in mem + hash checks. Will need to explore/evaluate.
- Compare perf and borrow learnings from other optimistic bridging protocols (nomad is the only one - https://www.nomad.xyz/)

Nomad's model is a `Home` contract computes a merkle root/commitment to all messages recorded there and optimistically broadcasts to `Replica` contracts. The `Replica` performs optimistic updating of the committed `Home` root. May be some 
https://github.com/nomad-xyz/monorepo/blob/9b6f68278cf687f9f19065d52b0fac6a128346b1/packages/contracts-core/contracts/Replica.sol#L122-L149

notbdu:
I think I might have used `hardhat` to profile during my proof verification testing but I don't think it's super granular. 

https://github.com/polymerdao/proof-verifier

contracts/OpConsensusStateManager.sol
zhengyangfeng00: I'll profile the gas cost, but want to clarify what you meant by shifting here. We are finding the smallest element, and then swapping it with the last element, and then pop the last element. it's less expensive than a real shift.

contracts/OpConsensusStateManager.sol
zhengyangfeng00: so far the only concern i have is about updating the fraud proof window. if updating fraud proof window is allowed, then an already trusted optimisitic consensus state may become pending again. need to think about it carefully to make sure that's not an issue.

zhengyangfeng00:
> Nomad's model is a `Home` contract computes a merkle root/commitment to all messages recorded there and optimistically broadcasts to `Replica` contracts. The `Replica` performs optimistic updating of the committed `Home` root. May be some learnings to borrow and a comparison point for perf.
> https://github.com/nomad-xyz/monorepo/blob/9b6f68278cf687f9f19065d52b0fac6a128346b1/packages/contracts-core/contracts/Replica.sol#L122-L149

thanks for the pointer! i'll compare with their approach and see what we can learn

contracts/OpConsensusStateManager.sol
notbdu: Oh woops didn't read properly. Yea this would be less expensive.

notbdu:
> > Nomad's model is a `Home` contract computes a merkle root/commitment to all messages recorded there and optimistically broadcasts to `Replica` contracts. The `Replica` performs optimistic updating of the committed `Home` root. May be some learnings to borrow and a comparison point for perf.
> > https://github.com/nomad-xyz/monorepo/blob/9b6f68278cf687f9f19065d52b0fac6a128346b1/packages/contracts-core/contracts/Replica.sol#L122-L149
> 
> thanks for the pointer! i'll compare with their approach and see what we can learn

Their approach is super simple AFAICT. Just storing a mapping of commitment root bytes to `confirmAt` time which is checked here:
https://github.com/nomad-xyz/monorepo/blob/9b6f68278cf687f9f19065d52b0fac6a128346b1/packages/contracts-core/contracts/Replica.sol#L263-L275

They then check that messages are only processed for _acceptable roots_. They do not trim any roots.

I have a feeling we can do even better than this based on a few observations:
- Deleting a storage slot refunds 15k gas
- The only `appHash` we care about keeping around is the latest _acceptable root_

alfredo-stonk:
PR looks good overall. 
The height of a `Proof` includes both a revision number (breaking changes in Polymer chain versions) and revision height (regular block height). If Polymer chain revision number changes, we would want to disable old revisions. We prob don't need to worry about that now though

notbdu (approved):
LGTM 🔥

PR #27, opened 2023-09-05
feat: optimize contract size with custom error

### LINKED ISSUE #13: Use custom error functions
Use custom error to save gas. 
Instead of `revert("error msg")`, do
```solidity

error invalidProof(); // no params
error InsufficientBalance(uint balance, uint withdrawAmount);

function testCustomError(uint _withdrawAmount) public view {
    uint bal = address(this).balance;
    if (bal < _withdrawAmount) {
        revert InsufficientBalance({balance: bal, withdrawAmount: _withdrawAmount});
    }
}
```

https://solidity-by-example.org/error/

zhengyangfeng00:
this change reduces the deployment size of the Dispatcher contract to below the threshold so that it is now deployable.

before the change:

Dispatcher contract size: 25.436, margin -0.86

after the change:

Dispatcher contract size: 24.332, margin 0.244

it fixes #13

### DIFF
--- a/contracts/Dispatcher.sol
+++ b/contracts/Dispatcher.sol
@@ -26,6 +26,39 @@ struct UpgradeClientMsg {
     ConsensusState consensusState;
 }
 
+// misc errors.
+error invalidCounterParty();
+error invalidCounterPartyPortId();
+error invalidHexStringLength();
+error invalidRelayerAddress();
+error consensusStateVerificationFailed();
+error packetNotTimedOut();
+
+// packet sequence related errors.
+error invalidPacketSequence();
+error unexpectedPacketSequence();
+
+// channel related errors.
+error channelNotOwnedBySender();
+error channelNotOwnedByPortAddress();
+
+// client related errors.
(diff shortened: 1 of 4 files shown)


nicopernas (approved):
Is there a way to provide a meaningful message along with the error? Just thinking about whomever gets the error back.

zhengyangfeng00:
> Is there a way to provide a meaningful message along with the error? Just thinking about whomever gets the error back.

ironically the example on ethereum doc site gives opposite example:

> Long revert messages and in particular many different revert messages can bloat up the contract. Instead use short error codes and decode them in your contract. A long message could be become much shorter:
>
> require(msg.sender == owner, "Only the owner of this contract can call this function");
>
> require(msg.sender == owner, "OW1");

source: https://ethereum.org/en/developers/tutorials/downsizing-contracts-to-fight-the-contract-size-limit/#shorten-error-message

i think because compute / storage resource is scarce in EVM, their suggestion almost means to give an arbitrary short and unique error identifier and let user read the source code to see what happens..

PR #29, opened 2023-09-25
refactor: change OpConsensusStateManager from library to contract

zhengyangfeng00:


nicopernas (approved):
What's benefit of having a "contract" as opposed to a library?

PR #33, opened 2023-11-14
add consensus state manager interface

nicopernas:


alfredo-stonk (approved):
Should we merge #21 too?

nicopernas:
> Should we merge #21 too?

That one went stale now. Need to update its version since main is ahead

PR #34, opened 2023-12-07
Universal channel - 1

alfredo-stonk:
Universal channel and IBC middleware implementation.

### What's done

- Universal Channel (UC) handshake
- Support full IBC packet semantics over UC
- Add IBC middleware (MW) contracts/interfaces which allows for composing arbitary MW stack
- Remove IBC fee and Escrow from Dispatcher
- Tests covering features above with `forge test`

Minor changes:
- using `forge test` Github Action, retiring `make test` which uses old Hardhat project setup
- dedup IBC events with IbcEventsEmitter interface
- add modifier functions for MW and IbcDispatcher callback access control
- add VirtualChain testing contracts which simulates separate deployment of contract instances on multiple virtual chains

Docs on Universal Channel and IBC Middleware
`docs/universal-channel-middleware.md` 

### How to test

```sh
forge test
```

Main test file is `test/universal.channel.t.sol`

Prerequisite:
Install Foundry at https://book.getfoundry.sh/getting-started/installation

### TODOs

- e2e test at polymerase repo
- 
Lower priorities
- add ics-29 packet fee MW 
- minor: migrate old Dispatcher tests to using VirtualChain for more test cases

PR #35, opened 2023-12-12
add openIbcChannel function signature to IbcDispatcher

tmsdkeys:
I've added the `openIbcChannel` function signature to the `IbcDispatcher` interface.

This enables the dApp to initiate a channel handshake from straight from the dApp, by calling the dispatcher **only having access to the IbcDispatcher ABI**. With current design, user would need to import entire Dispatcher ABI.

Mars contract has been updated with an example function `initiateChannelHandshake`

### DIFF
--- a/contracts/Mars.sol
+++ b/contracts/Mars.sol
@@ -112,6 +112,19 @@ contract Mars is IbcReceiver, Ownable {
         dispatcher.closeIbcChannel(channelId);
     }
 
+    function initiateChannelHandshake(
+        IbcDispatcher dispatcher,
+        IbcReceiver portAddress,
+        string calldata version,
+        ChannelOrder ordering,
+        bool feeEnabled,
+        string[] calldata connectionHops,
+        CounterParty calldata counterparty,
+        Proof calldata proof
+    ) external {
+        dispatcher.openIbcChannel(portAddress, version, ordering, feeEnabled,connectionHops, counterparty, proof);
+    }    
+
     function greet(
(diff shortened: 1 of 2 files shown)


tmsdkeys:
@alfredo-stonk should I add a fee here as well?

contracts/Mars.sol
```diff
@@ -112,6 +112,19 @@ contract Mars is IbcReceiver, Ownable {
...
+        IbcDispatcher dispatcher,
+        IbcReceiver portAddress,
+        string calldata version,
+        ChannelOrder ordering,
+        bool feeEnabled,
+        string[] calldata connectionHops,
+        CounterParty calldata counterparty,
+        Proof calldata proof
+    ) external {
+        dispatcher.openIbcChannel(portAddress, version, ordering, feeEnabled,connectionHops, counterparty, proof);
```
tmsdkeys, line 125: @alfredo-stonk should I add a fee here as well?

alfredo-stonk: No fee for channel handshake while calling Dispatcher.  
We also removed packet fee processing from Dispatcher too, as it's handled by IBC middleware. Lots of refactoring to support general MW stack in #34. 

If dApps need a ics-29 fee-enabled channel, they'd call fee-MW's `openIbcChannel` func.

Let's deal no-fee only channel in this PR for now? We're prioritizing Catalyst's GeneralisedIncentive MW

tmsdkeys:
Implemented in #34

PR #41, opened 2024-02-09
copy functions from Ibc.sol to dispatcher to workaround a few issues

zhengyangfeng00:
## Background

This PR addresses some problems encountered during the deployment of contracts.

In the file `Ibc.sol`, there's a library called Ibc which contains both public and internal functions. When a library includes public functions, it can't be directly included and must be linked separately during the contract compilation process. This makes deploying contracts more complicated. Additionally, the library we used for deploying contracts doesn't handle linked libraries well, especially when it comes to recognizing enum types(see https://github.com/dethcrypto/TypeChain/issues/216).

A somewhat hacky solution(https://github.com/dethcrypto/TypeChain/pull/281) was added to the typechain package to fix the enum types issue when using linked libraries. However, this solution doesn't cover the ethersjs-v6 target, which is used in our end-to-end test server.

There are two options to fix this problem:

1. Convert Ibc from a library to a separate contract. This might work, but it would mean that all contracts using the utility functions would need an extra step of linking during compilation.
2. Copy relevant functions from the Ibc library to a dispatcher contract and remove the import of Ibc.sol from the dispatcher contract.

This PR has chosen the second option because it's the simplest solution.

## Test

1. update the vibc-core-sc commit to this one in polymerase repo
2. start the test server and trigger a deployment with gql query
3. the server output showed that the contracts are deployed successfully

```gql
mutation {
  evm(chainName: "eth1") {
    deployPolymerContracts(force: false) {
      deployedContracts {
        name
        address
      }
    }
  }
}
```

```
❯ task start-server-dev 2>&1 | tee server.log
task: Task "build-peptide" is up to date
task: Task "build-oprelayer" is up to date
task: Task "build-proofservice" is up to date
task: [start-server-dev] bun run bin/server.ts start -c config/server.config.yaml --wd /Users/zhengyangfeng/polymerase/tests/e2e/__private --start-chains 'eth1,eth2'
Server is running on http://localhost:4000/graphql
JsonRpcProvider failed to detect network and cannot start up; retry in 1s (perhaps the URL is wrong or the node is not started)
JsonRpcProvider failed to detect network and cannot start up; retry in 1s (perhaps the URL is wrong or the node is not started)
chain [eth1] is live with blockNum = 0
chain [eth2] is live with blockNum = 0
[eth1] deploying Verifier
[eth1] deploying ConsensusStateManager
[eth1] deploying Verifier...Done
[eth1] deploying ConsensusStateManager...Done
[eth1] deploying Dispatcher
[eth1] deploying Dispatcher...Done
[eth1] deploying Mars
[eth1] deploying UniversalChannelHandler
[eth1] deploying Mars...Done
[eth1] deploying UniversalChannelHandler...Done
[eth1] deploying Earth
[eth1] deploying Earth...Done
```


zhengyangfeng00:
no longer needed since the original pr has been updated

PR #47, opened 2024-02-29
Raunak/multi client support

RnkSngh:
I took a cursory stab at:

- reducing contract size (through exporting functions to the `IbcUtils` library (we might still be just a hair above the contract size limit, but this can be solved by playing around with the solidity compiler) 
- renamed `consensusManager` -> `lightClient` as per @alfredo-stonk 's suggestion
- Added a mapping for getting a client id from a connectionId (assuming a client id will just start at 0, and increment from there? ), and also added a corresponding setter

I'm still wondering what the best way is to get the clientId to the `timeout`, `acknowledgement`, and `recvPacket` methods - these all require proving from a light client. Right now I just added an extra argument to all of these methods called `clientId`, though I'm not sure if this goes against the [IC4 spec](https://github.com/cosmos/ibc/blob/main/spec/core/ics-004-channel-and-packet-semantics/README.md#writing-acknowledgements) since that doesn't seem to give any client identifiers, or if this would make life hard for integrating these changes to our sequencer. From my understanding the portId contains the clientId as well, and each of these functions take a `IbcPacket` type, which has some port data associated with it - I think @alfredo-stonk said he might be thinking about what to do with the portID and how to decouple that from the dispatcher contract so that might also be relevant for this decision

Another thing I need to check is how to access the connection from the `connectionHops` array arguments given to `openIbcChannel`, `verifyConnectIbcChannelProof` , `connectIbcChannel` (right now I'm just taking the first index at each but need to understand this deeper)

I also still need to write unit tests in forge! and then after that will be the Sepolia deployment

### DIFF
--- a/contracts/core/Dispatcher.sol
+++ b/contracts/core/Dispatcher.sol
@@ -35,6 +35,8 @@ contract Dispatcher is IbcDispatcher, IbcEventsEmitter, Ownable, Ibc {
     string public portPrefix;
     uint32 portPrefixLen; // TO DO: Will need to change this to reflect dispatchers having multiple clients
 
+    mapping(string => uint256) public connectionIdToClientId;
+    mapping(uint256 => LightClient) public clientIdToLightClient;
     mapping(address => mapping(bytes32 => Channel)) public portChannelMap;
     mapping(address => mapping(bytes32 => uint64)) nextSequenceSend;
     // keep track of received packets' sequences to ensure channel ordering is enforced for ordered channels
@@ -49,15 +51,12 @@ contract Dispatcher is IbcDispatcher, IbcEventsEmitter, Ownable, Ibc {
     // keep track of outbound ack packets to prevent replay attack
     mapping(address => mapping(bytes32 => mapping(uint64 => bool))) ackPacketCommitment;
 
-    LightClient lightClient;
-
     //
     // methods
     //
-    constructor(string memory initPortPrefix, LightClient _lightClient) {
+    constructor(string memory initPortPrefix) {
         portPrefix = initPortPrefix;
         portPrefixLen = uint32(bytes(initPortPrefix).length);
-        lightClient = _lightClient;
     }
 
     //
@@ -90,17 +89,28 @@ contract Dispatcher is IbcDispatcher, IbcEventsEmitter, Ownable, Ibc {
         L1Header calldata l1header,
         OpL2StateProof calldata proof,
         uint256 height,
-        uint256 appHash
+        uint256 appHash,
+        uint256 clientId
     ) external returns (uint256 fraudProofEndTime, bool ended) {
+        LightClient lightClient = clientIdToLightClient[clientId];
+        if (address(lightClient) == address(0)) {
+            revert IBCErrors.clientNotFound();
+        }
+
         return lightClient.addOpConsensusState(l1header, proof, height, appHash);
     }
 
     // getOptimisticConsensusState
-    function getOptimisticConsensusState(uint256 height)
+    function getOptimisticConsensusState(uint256 height, uint256 clientId)
         external
         view
         returns (uint256 appHash, uint256 fraudProofEndTime, bool ended)
     {
+        LightClient lightClient = clientIdToLightClient[clientId];
+        if (address(lightClient) == address(0)) {
+            revert IBCErrors.clientNotFound();
+        }
+
         return lightClient.getState(height);
     }
 
@@ -129,6 +139,11 @@ contract Dispatcher is IbcDispatcher, IbcEventsEmitter, Ownable, Ibc {
         }
 
         if (IbcUtils.isChannelOpenTry(counterparty)) {
+            uint256 clientId = connectionIdToClientId[connectionHops[0]];
+            LightClient lightClient = clientIdToLightClient[clientId];
+            if (address(lightClient) == address(0) || clientId == 0) {
+                revert IBCErrors.clientNotFound();
+            }
             lightClient.verifyMembership(
                 proof,
                 channelProofKey(local.portId, local.channelId),
@@ -158,6 +173,11 @@ contract Dispatcher is IbcDispatcher, IbcEventsEmitter, Ownable, Ibc {
         CounterParty calldata counterparty,
         Ics23Proof calldata proof
     ) internal {
+        uint256 clientId = connectionIdToClientId[connectionHops[0]];
+        LightClient lightClient = clientIdToLightClient[clientId];
+        if (address(lightClient) == address(0) || clientId == 0) {
+            revert IBCErrors.clientNotFound();
+        }
         lightClient.verifyMembership(
             proof,
             channelProofKey(local.portId, local.channelId),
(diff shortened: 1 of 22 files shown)


alfredo-stonk:
Good first steps 👍 

How to get client from connectionHops during channel handshakes?
- take the first connectionId and look it up in the local connectionToClientMap

How to get client from packet?
- take source channelId field from `IbcPacket`, map it to connectionId, then to clientId

Re. portPreifx, after some thought, we prob won't need to change it. That is, one portPrefix per virtual chain deployment as is now. 
- Reasoning is the same as core IBC specs. Ports are just static addresses. They don't carry connection topology, unlike conenctions/channels. One port is expected to work with multiple clients/connections/channels

RnkSngh:
> take source channelId field from IbcPacket, map it to connectionId, then to clientId

Ah I see, so the channelId to connectionId mapping ( I don't think one currently exists on the contract?) will probably also need to be set in a setter by the relayer? Similar to how now we added another setter for mapping connectionIds to clientIds?

contracts/core/OpLightClient.sol
```diff
@@ -40,11 +39,7 @@ contract OptimisticConsensusStateManager is ConsensusStateManager {
             // if this is a new apphash we need to verify the provided proof. This method will revert in case
             // of invalid proof.
             verifier.verifyStateUpdate(
-                l1header,
-                proof,
-                bytes32(appHash),
-                l1BlockProvider.hash(),
-                l1BlockProvider.number()
+                l1header, proof, bytes32(appHash), l1BlockProvider.hash(), l1BlockProvider.number()
```
nicopernas, line 42: There's a `.prettierrc.yml` config in the root directory that specifies minor formatting config that we all follow. You may wanna add that to your IDE to prevent changes like this.

contracts/interfaces/IbcMiddleware.sol
```diff
@@ -220,7 +212,7 @@ contract IbcMwUser is Ownable {
      * Should add this modifier to all IBC-related callback functions.
      */
     modifier onlyIbcMw() {
-        require(authorizedMws[msg.sender], 'unauthorized IBC middleware');
+        require(authorizedMws[msg.sender], "unauthorized IBC middleware");
```
nicopernas, line 215: This is also something that `.prettierrc.yml` takes care of

contracts/core/Dispatcher.sol
```diff
@@ -37,64 +51,26 @@ contract Dispatcher is IbcDispatcher, IbcEventsEmitter, Ownable, Ibc {
...
-
     // verify an EVM address matches an IBC portId.
     // IBC_PortID = portPrefix + address (hex string without 0x prefix, case-insensitive)
     function portIdAddressMatch(address addr, string calldata portId) public view returns (bool) {
         if (keccak256(abi.encodePacked(portPrefix)) != keccak256(abi.encodePacked(portId[0:portPrefixLen]))) {
             return false;
         }
         string memory portSuffix = portId[portPrefixLen:];
-        return hexStrToAddress(portSuffix) == addr;
+        return IbcUtils.hexStrToAddress(portSuffix) == addr;
```
nicopernas, line 73: One of the reasons we didn't use libraries to their full extent is because we ran into troubles when generating Typescript bindings (https://github.com/open-ibc/vibc-core-smart-contracts/pull/41) 
At the time, the simplest solution was to avoid them in some places, that's why you'll see that the `Dispatcher` contract inherits from `Ibc` as a opposed to use it as a library.  In any case, I think libraries make the most sense and we should fix or workaround the binding issues. 

Just wanted to call this out in case you run into the same problems. And if so, now is the time to find a real fix.

contracts/core/Dispatcher.sol
```diff
@@ -173,24 +135,24 @@ contract Dispatcher is IbcDispatcher, IbcEventsEmitter, Ownable, Ibc {
         Ics23Proof calldata proof
     ) external {
         if (bytes(counterparty.portId).length == 0) {
-            revert invalidCounterPartyPortId();
+            revert IBCErrors.invalidCounterPartyPortId();
         }
 
-        if (isChannelOpenTry(counterparty)) {
-            consensusStateManager.verifyMembership(
+        if (IbcUtils.isChannelOpenTry(counterparty)) {
```
nicopernas, line 141: I am currently working on an API clean up that will remove this method and change the channel handshake api a bit. Maybe leave it here instead of moving it to a library to avoid bigger conflicts?

contracts/core/Dispatcher.sol
```diff
@@ -414,20 +387,26 @@ contract Dispatcher is IbcDispatcher, IbcEventsEmitter, Ownable, Ibc {
...
+            revert IBCErrors.receiverNotIndtendedPacketDestination();
         }
 
         // prove absence of packet receipt on Polymer chain
         // TODO: add non membership support
-        consensusStateManager.verifyNonMembership(proof, 'packet/receipt/path');
+        LightClient lightClient = clientIdToLightClient[clientId];
+        if (address(lightClient) == address(0) || clientId == 0) {
+            revert IBCErrors.clientNotFound();
+        }
```
nicopernas, line 403: this code seems to be used in different places. Maybe we wanna move it to its own function or library?

contracts/core/Dispatcher.sol
```diff
@@ -438,6 +417,25 @@ contract Dispatcher is IbcDispatcher, IbcEventsEmitter, Ownable, Ibc {
...
+        if (clientId == 0 || bytes(connectionId).length == 0) {
+            revert IBCErrors.invalidInput();
+        }
+
+        if (connectionIdToClientId[connectionId] != 0 || clientIdToLightClient[clientId] != LightClient(address(0))) {
+            revert IBCErrors.invalidInput();
+        }
+
+        connectionIdToClientId[connectionId] = clientId;
+        clientIdToLightClient[clientId] = newLightClient;
```
nicopernas, line 436: As per Alexander's comments on https://github.com/open-ibc/vibc-core-smart-contracts/pull/48 it seems like having a single mapping of complex types is more efficient than multiple maps of simple types. An option will be to merge these two maps into one

contracts/core/Dispatcher.sol
RnkSngh: Yes! in this branch we were running up against the contract size limit - so I didn't abstract these into different functions (IIRC adding functions increases the contract size); they can't be a library because these require access to state and Libraries can't hold stated.

But if we can change this code use a library instead of inherit from it and get the typechain integration working, we can definitely clean this up

contracts/core/Dispatcher.sol
RnkSngh: yes! Planning to implement this in the Proxy design PR

RnkSngh:
I paused development on this PR for now; since [the CI PR](https://github.com/open-ibc/vibc-core-smart-contracts/pull/53) and [the upgradeable proxy branch](https://github.com/open-ibc/vibc-core-smart-contracts/tree/raunak/upgradeable-dispatcher) will likely get merged before this one

contracts/core/Dispatcher.sol
reednaa: That is only true when the struct can be packed in fewer storage slots. For example, in the other case you were using 3x1 byte (3 bools) which obviously fit in 32 bytes.
I am not sure these can be compressed further.

PR #48, opened 2024-02-29
Gas optimisations

reednaa:
Save ~85564 gas across function calls in the test suite. Most notisable improvements are found in Dispatcher.sol, with an average of 7% cost reductions with `recvPacket` seeing an 11% gas reduction.

For an overview of difference between gas reports, see:
https://docs.google.com/spreadsheets/d/1GHleuGHfqNIN7fHs8T7KiOKs8WH-fdGElh0ysYavanY/edit?usp=sharing

There are more savings to be hand with unchecked, however, I couldn't figure out the strict bonds. Over all: The implementation is already pretty optimised.

## Summary by CodeRabbit

- **New Features**
	- Introduced new data structures for packet sequences and commitments in the Dispatcher contract.
	- Added `immutable` modifiers to critical variables across contracts for enhanced security and efficiency.
- **Refactor**
	- Improved data construction for packet transmission using `bytes.concat` in multiple contracts.
	- Optimized loop iterations in the Universal Channel Handler and Ibc contracts for better performance.
	- Reorganized mappings and updated constructor in the Dispatcher contract for clarity and functionality.
- **Bug Fixes**
	- Fixed loop progression issue in the Universal Channel Handler contract.
- **Tests**
	- Adjusted function signature in Dispatcher contract tests for consistency.

### DIFF
--- a/contracts/core/Dispatcher.sol
+++ b/contracts/core/Dispatcher.sol
@@ -520,8 +520,7 @@ contract Dispatcher is IbcDispatcher, IbcEventsEmitter, Ownable, Ibc {
         }
 
         // Not timeout yet, then do normal handling
-        IbcPacket memory pkt = packet;
-        AckPacket memory ack = receiver.onRecvPacket(pkt);
+        AckPacket memory ack = receiver.onRecvPacket(packet);
         bool hasAckPacketCommitment = packetCommitment[address(receiver)][packet.dest.channelId][packet.sequence].ackPacked;
         // check is not necessary for sync-acks
         if (hasAckPacketCommitment) {
--- a/contracts/core/Dispatcher.sol
+++ b/contracts/core/Dispatcher.sol
@@ -8,6 +8,18 @@ import '../interfaces/IbcDispatcher.sol';
 import {IbcChannelReceiver, IbcPacketReceiver} from '../interfaces/IbcReceiver.sol';
 import '../interfaces/ConsensusStateManager.sol';
 
+struct PackedNextSequence {
+    uint64 send;
+    uint64 recv;
+    uint64 ack;
+}
+
+struct PackedCommits {
+    bool sendPacked;
+    bool recvPacked;
+    bool ackPacked;
+}
+
 /**
  * @title Dispatcher
  * @author Polymer Labs
@@ -24,25 +36,19 @@ contract Dispatcher is IbcDispatcher, IbcEventsEmitter, Ownable, Ibc {
     uint32 portPrefixLen;
 
     mapping(address => mapping(bytes32 => Channel)) public portChannelMap;
-    mapping(address => mapping(bytes32 => uint64)) nextSequenceSend;
     // keep track of received packets' sequences to ensure channel ordering is enforced for ordered channels
-    mapping(address => mapping(bytes32 => uint64)) nextSequenceRecv;
-    mapping(address => mapping(bytes32 => uint64)) nextSequenceAck;
+    mapping(address => mapping(bytes32 => PackedNextSequence)) nextSequence;
     // only stores a bit to mark packet has not been ack'ed or timed out yet; actual IBC packet verification is done on
     // Polymer chain.
-    // Keep track of sent packets
-    mapping(address => mapping(bytes32 => mapping(uint64 => bool))) public sendPacketCommitment;
-    // keep track of received packets to prevent replay attack
-    mapping(address => mapping(bytes32 => mapping(uint64 => bool))) recvPacketReceipt;
-    // keep track of outbound ack packets to prevent replay attack
-    mapping(address => mapping(bytes32 => mapping(uint64 => bool))) ackPacketCommitment;
+    // keep track of packets to prevent replay attack
+    mapping(address => mapping(bytes32 => mapping(uint64 => PackedCommits))) public packetCommitment;
 
     ConsensusStateManager consensusStateManager;
 
     //
     // methods
     //
-    constructor(string memory initPortPrefix, ConsensusStateManager _consensusStateManager) {
+    constructor(string memory initPortPrefix, ConsensusStateManager _consensusStateManager) payable {
         portPrefix = initPortPrefix;
         portPrefixLen = uint32(bytes(initPortPrefix).length);
         consensusStateManager = _consensusStateManager;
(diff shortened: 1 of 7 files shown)


contracts/core/Dispatcher.sol
```diff
@@ -65,18 +71,27 @@ contract Dispatcher is IbcDispatcher, IbcEventsEmitter, Ownable, Ibc {
...
-            uint8 high = uint8(strBytes[i * 2]);
-            uint8 low = uint8(strBytes[1 + i * 2]);
-            // Convert to lowercase if the character is in uppercase
-            if (high >= 65 && high <= 90) {
-                high += 32;
-            }
-            if (low >= 65 && low <= 90) {
-                low += 32;
+
+        for (uint256 i = 0; i < 20;) {
```
RnkSngh, line 75: I think these bounds are all correct, but in this specific instance, I'm also slightly leaning towards "this unchecked math is too complicated for me to be absolutely sure that there isn't an over/underflow risk". 

Especially for L2s, I think the gas savings here would be negligible compared to the additional under/overflow risk added here. 

But also, I want to at some point take a stab at replacing this method completely with the `toHexStr` method in the OZ's strings library as well. Though it seemed like @nicopernas had imported that library in some parts of this codebase so maybe he has already tried that and it's not useful here.

contracts/core/Dispatcher.sol
```diff
@@ -8,6 +8,18 @@ import '../interfaces/IbcDispatcher.sol';
 import {IbcChannelReceiver, IbcPacketReceiver} from '../interfaces/IbcReceiver.sol';
 import '../interfaces/ConsensusStateManager.sol';
 
+struct PackedNextSequence {
+    uint64 send;
```
RnkSngh, line 12: This struct packing is brilliant, thanks for this @reednaa !

contracts/core/Dispatcher.sol
```diff
@@ -390,17 +411,19 @@ contract Dispatcher is IbcDispatcher, IbcEventsEmitter, Ownable, Ibc {
...
 
         if (channel.ordering == ChannelOrder.ORDERED) {
-            if (packet.sequence != nextSequenceAck[address(receiver)][packet.src.channelId]) {
+            if (packet.sequence != nextSequence[address(receiver)][packet.src.channelId].ack) {
                 revert unexpectedPacketSequence();
             }
 
-            nextSequenceAck[address(receiver)][packet.src.channelId] = packet.sequence + 1;
+            unchecked {
+                nextSequence[address(receiver)][packet.src.channelId].ack = packet.sequence + 1;
```
RnkSngh, line 419: Since the `packet` is sent directly through calldata, could there be problems with some inputs? If there's a packet with the sequence set to `max(uint64)`, this would result in an overflow. 

I'm leaning towards suggesting to leave this as checked math for now since I'm not sure what the implications are for an overflow happening here

contracts/core/Dispatcher.sol
```diff
@@ -457,25 +480,27 @@ contract Dispatcher is IbcDispatcher, IbcEventsEmitter, Ownable, Ibc {
...
         Channel memory channel = portChannelMap[address(receiver)][packet.dest.channelId];
         if (channel.ordering == ChannelOrder.ORDERED) {
-            if (packet.sequence != nextSequenceRecv[address(receiver)][packet.dest.channelId]) {
+            if (packet.sequence != nextSequence[address(receiver)][packet.dest.channelId].recv) {
                 revert unexpectedPacketSequence();
             }
 
-            nextSequenceRecv[address(receiver)][packet.dest.channelId] = packet.sequence + 1;
+            unchecked {
+                nextSequence[address(receiver)][packet.dest.channelId].recv = packet.sequence + 1;
```
RnkSngh, line 502: see other comment about potential overflow in the `acknowledgement` method

contracts/libs/Ibc.sol
```diff
@@ -213,11 +217,16 @@ contract Ibc {
...
-            i++;
+            unchecked {
+                ++i;
+            }
         }
         bytes memory bytesArray = new bytes(i);
-        for (uint8 j = 0; j < i; j++) {
+        for (uint8 j = 0; j < i;) {
             bytesArray[j] = b[j];
+            unchecked {
```
RnkSngh, line 227: I think this unchecked math is fine for now, though we might also want to look into using OZ's string util methods here as well

RnkSngh (commented):
Hi Alexander, thanks for taking a look and for running the gas analysis! 

I really appreciated your suggestion on packing the send/recv/ack into one struct to simplify the mapping; that considerably simplifies some things! 

In general I agree with most of your unchecked math except for the couple of spots where I left comments

contracts/core/Dispatcher.sol
```diff
@@ -24,25 +36,19 @@ contract Dispatcher is IbcDispatcher, IbcEventsEmitter, Ownable, Ibc {
...
+    // keep track of packets to prevent replay attack
+    mapping(address => mapping(bytes32 => mapping(uint64 => PackedCommits))) public packetCommitment;
 
     ConsensusStateManager consensusStateManager;
 
     //
     // methods
     //
-    constructor(string memory initPortPrefix, ConsensusStateManager _consensusStateManager) {
+    constructor(string memory initPortPrefix, ConsensusStateManager _consensusStateManager) payable {
```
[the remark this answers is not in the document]
reednaa: If there is no payable here, Solidity adds a check for `msg.value == 0`. It doesn't have any functionality expect add excess logic (size) and a very small cost.

contracts/core/Dispatcher.sol
reednaa: Note that this doesn't actually save as much gas as you might imagine. The slots are used independently of each other, so you don't save on warming up the slots which often contribute the biggest amount.
However, you do still save a lot on only changing a 0 slot to non-zero 1 time instead of 3 times. That is (if I recall correctly) roughly 2x~20k gas saves.

contracts/core/Dispatcher.sol
reednaa: I assumed that this check: `(packet.sequence != nextSequence[address(receiver)][packet.src.channelId].ack)` bounded it.

contracts/core/Dispatcher.sol
reednaa: I assmued that this check: `if (packet.sequence != nextSequence[address(receiver)][packet.dest.channelId].recv)` bounded it.

contracts/libs/Ibc.sol
reednaa: The argument is that j is bounded by i which is bounded by 32.

contracts/core/Dispatcher.sol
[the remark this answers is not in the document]
RnkSngh: Ah gotcha; this makes sense!

contracts/core/Dispatcher.sol
RnkSngh: Ah ok, I think this is valid

contracts/core/Dispatcher.sol
RnkSngh, line 51:
```suggestion
    constructor(string memory initPortPrefix, ConsensusStateManager _consensusStateManager) {
```

Actually, I think we should leave this payable out because the solidity check that msg.value = 0 is useful for ensuring that we don't accidentally send any tokens to this during deployment. I'm mainly thinking of the case where something is wrong in our deployment script and we accidentally pass in a value for this consturctor deployment - all the tokens would be lost in this contract if that happens. Granted this has a very small likelyhood of happening, but imo it's worth the small gas cost.

Also this gas optimization would only impact deployment cost from our end, and not dapp devs who are using this so I think it's fine for us to pay this additional gas cost to insure we don't send any tokens to this contract that would be lost.

RnkSngh:
Hey @reednaa , after a discussion with @dshiell, we decided we should hold off on the more complicated unchecked math, mainly because it's not quite worth the over/underflow potential, and the cost reductions aren't a ton considering we're first deploying on L2s. I also think we'll soon refactor this code to use OZ's string libaries anyways. 

If you can get rid of the unchecked math (I think for the for loops where it's probably fine), and also address my new constructor payable comment - I can approve this PR!

dshiell:
@reednaa - What does the gas reduction percent look like without the unchecked math? If we still get the majority of the efficiency from other changes then maybe we can split the unchecked math optimizations into another PR to reduce any risks in the short term while still gaining most of the efficiency.

reednaa:
They shouldn't be significant. The majority of the savings comes from immutable storage and the boolean struct.

RnkSngh:
closing this PR for now, will have gas optimizations very soon!

PR #52, opened 2024-03-01
CI 

RnkSngh:
PR to add listing and formatting to our CI (changes to actual solidity code were just those automatically created using `forge fmt` and should only include changes to white space or converting single quotes to double quotes. 

Right now I'm just using the standard settings that come with the forge formatter, but this can be configured in the `foundry.toml` [here](https://book.getfoundry.sh/reference/config/formatter). If anyone has strong opinions on whitespace, single vs double quotes etc - speak now or forever hold your peace!

I will set up solhint + forge fmt check tomorrow morning!


RnkSngh:
Had some trouble with diverging git histories - I think simplest solution is to close this one & open a pr from another branch [here](https://github.com/open-ibc/vibc-core-smart-contracts/pull/53)

PR #55, opened 2024-03-04
nico/new channel api

nicopernas:
Split the OpenIbcChannel and ConnectIbcChannel apis into their correspondent Ibc channel handshake
methods:

- ChannelOpenInit
- ChannelOpenTry
- ChannelOpenAck
- ChannelOpenConfirm

## Summary by CodeRabbit

- **Refactor**
	- Updated IBC channel opening and connection logic in multiple contracts by renaming functions and events for clarity and consistency.
- **Bug Fixes**
	- Streamlined version handling in IBC channel interactions by removing the `VersionMismatch` error.
- **Tests**
	- Adjusted and renamed test functions to match the refactored IBC channel opening and connection logic for accurate testing.

### DIFF
--- a/contracts/core/Dispatcher.sol
+++ b/contracts/core/Dispatcher.sol
@@ -68,13 +68,35 @@ contract Dispatcher is IbcDispatcher, IbcEventsEmitter, Ownable, Ibc {
     }
 
     /**
-     * This func is called by a 'relayer' on behalf of a dApp. The dApp should be implements IbcChannelHandler.
-     * The dApp should implement the onOpenIbcChannel method to handle one of the first two channel handshake methods,
-     * ie. ChanOpenInit or ChanOpenTry.
-     * If callback succeeds, the dApp should return the selected version, and an emitted event will be relayed to the
-     * IBC/VIBC hub chain.
+     * This function is called by a 'relayer' on behalf of a dApp. The dApp should implement IbcChannelHandler's
+     * onChanOpenInit. If the callback succeeds, the dApp should return the selected version and the emitted event
+     * will be relayed to the  IBC/VIBC hub chain.
      */
-    function openIbcChannel(
+    function channelOpenInit(
+        IbcChannelReceiver portAddress,
+        string calldata version,
+        ChannelOrder ordering,
+        bool feeEnabled,
+        string[] calldata connectionHops,
+        string calldata counterpartyPortId
+    ) external {
+        if (bytes(counterpartyPortId).length == 0) {
+            revert IBCErrors.invalidCounterPartyPortId();
+        }
+
+        string memory selectedVersion = portAddress.onChanOpenInit(version);
+
+        emit ChannelOpenInit(
+            address(portAddress), selectedVersion, ordering, feeEnabled, connectionHops, counterpartyPortId
+        );
+    }
+
+    /**
+     * This function is called by a 'relayer' on behalf of a dApp. The dApp should implement IbcChannelHandler's
+     * onChanOpenTry. If the callback succeeds, the dApp should return the selected version and the emitted event
+     * will be relayed to the  IBC/VIBC hub chain.
+     */
+    function channelOpenTry(
         IbcChannelReceiver portAddress,
         CounterParty calldata local,
         ChannelOrder ordering,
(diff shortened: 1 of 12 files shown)


contracts/interfaces/IbcReceiver.sol
```diff
@@ -12,16 +12,11 @@ import {ChannelOrder, CounterParty, IbcPacket, AckPacket} from "../libs/Ibc.sol"
...
-        ChannelOrder ordering,
-        bool feeEnabled,
-        string[] calldata connectionHops,
-        CounterParty calldata counterparty
-    ) external returns (string memory selectedVersion);
+    function onChannelOpenInit(string calldata version) external returns (string memory selectedVersion);
 
-    function onConnectIbcChannel(bytes32 channelId, bytes32 counterpartyChannelId, string calldata counterpartyVersion)
-        external;
+    function onChannelOpenTry(string calldata counterpartyVersion) external returns (string memory selectedVersion);
```
nicopernas, line 17: At this point the dapp is only responsible for the version selection and it seems like providing the version and counterparty version is enough. Something like this
- initiator sets local version to X 
- initiator dapp checks that X is valid, so it becomes the "selected version"
- on remote chain, channOpenTry is called with counterpartyVersion = X
- remote dapp is called with counterparty version X and if it supports it, it becomes the selected version. 
I think this is cleaner than having to compare local with remote version

contracts/interfaces/IbcReceiver.sol
```diff
@@ -12,16 +12,11 @@ import {ChannelOrder, CounterParty, IbcPacket, AckPacket} from "../libs/Ibc.sol"
...
+
+    function onChannelConnect(bytes32 channelId, string calldata counterpartyVersion) external;
```
nicopernas, line 19: this one did not seem like it would benefit from a split so I left the same callback for both ack and confirm steps. Open to other options though.

contracts/core/Dispatcher.sol
```diff
@@ -453,25 +481,33 @@ contract Dispatcher is IbcDispatcher, IbcEventsEmitter, Ownable, Ibc {
...
-            proof,
-            channelProofKey(local.portId, local.channelId),
-            channelProofValue(
-                isChanConfirm ? ChannelState.CONFIRM_PENDING : ChannelState.ACK_PENDING,
-                ordering,
-                local.version,
-                connectionHops,
-                counterparty
-            )
+        portAddress.onChannelConnect(local.channelId, counterparty.version);
```
RnkSngh, line 492: These callbacks have slightly different naming than the one in the [Polymer Docs here](https://docs.polymerlabs.org/docs/learn/concepts/ibc/#channels-ics-4) - not sure if this is something we should change in this PR (considering that there are already other interface changes) or in another PR or if this is something we want to change at all

contracts/core/Dispatcher.sol
```diff
@@ -453,25 +481,33 @@ contract Dispatcher is IbcDispatcher, IbcEventsEmitter, Ownable, Ibc {
...
+
```
RnkSngh, line 493: Consider moving the `onChannelConnect` call to the end of this function so as to follow the [checks-effects-interactions](https://fravoll.github.io/solidity-patterns/checks_effects_interactions.html) pattern. 

I don't think this particularly insecure code if someone _were_ to reenter this contract, but in general it is always good to follow this pattern to prevent random code injections that can come from recursively calling this function in the `onChannelConnect` handler

contracts/interfaces/IbcReceiver.sol
RnkSngh: how would the initiator dapp be notified if the remote chain doesn't support the version - I'm assuming that wouldn't be called in a handler right? So would it just timeout?

RnkSngh (commented):
These changes look good! This already seems simpler on the smart contract level; and I'm sure this will further simplify the rollup code as well!

contracts/core/Dispatcher.sol
nicopernas: I asked that question myself and thought it would be easier to have the one `onChannelConnect` instead of splitting it in two. But given the docs you pointed out, I think you have a valid point. Let me split them too and be consistent throughout

contracts/interfaces/IbcReceiver.sol
nicopernas: The remote chain will error out and the channel will never be established. Channel handshake messages don't really timeout as packet messages do. The channel will stay in INIT state

contracts/interfaces/IbcReceiver.sol
nicopernas: will finally split as mentioned in previous comment so we are more consistent.

contracts/core/Dispatcher.sol
RnkSngh: awesome, thank you for implementing this!

RnkSngh (approved):

PR #56, opened 2024-03-06
Raunak/upgradeable dispatcher

RnkSngh:
This PR makes the Dispatcher an upgradeable proxy, and a corresponding test suite. This should hopefully make deployments less of a headache. Let me know if there are any questions around security/upgradeability!


RnkSngh:
Dispatcher should be deployed! Here:

- [Proxy](https://optimism-sepolia.blockscout.com/address/0x20d7Bd73A52d3522Dfa38D4a458A23a8F4472381?tab=contract)
- [Implementation](0x9AB57459edD3B005F1B81bEB9273c5E10c7f34A1)

dshiell (approved):
approved, contingent on deciding on a stable contract storage layout. Is there somewhere we need to put a big warning about not changing contract parameters and only adding?

test/Dispatcher.proof.t.sol
```diff
@@ -70,8 +66,11 @@ contract DispatcherIbcWithRealProofs is IbcEventsEmitter, ProofBase {
         Ics23Proof memory proof = load_proof("/test/payload/packet_ack_proof.hex");
 
         // plant a fake packet commitment so the ack checks go through
-        stdstore.target(address(dispatcher)).sig(dispatcher.sendPacketCommitment.selector).with_key(address(mars))
-            .with_key(ch0.channelId).with_key(uint256(1)).checked_write(true);
+        // Stdstore doesn't work for proxies so we have to use store
+        bytes32 slot1 = keccak256(abi.encode(address(mars), uint32(107))); // current nested mapping slot: 107
```
nicopernas, line 70: how did you calculate the 107?

test/Dispatcher.t.sol
```diff
@@ -4,23 +4,18 @@ pragma solidity ^0.8.13;
...
+import {IbcReceiver} from "../contracts/interfaces/IbcReceiver.sol";
+import {DummyLightClient} from "../contracts/utils/DummyLightClient.sol";
+import {TestUtils} from "./TestUtils.sol";
 import "../contracts/examples/Mars.sol";
-import "../contracts/core/OpConsensusStateManager.sol";
+import "../contracts/core/OpLightClient.sol";
 import "./Dispatcher.base.t.sol";
 
-contract ChannelHandshakeTest is Base {
+abstract contract ChannelHandshakeTestSuite is Base {
```
nicopernas, line 14: why do we new the `abstract` keyword here? are we missing implementation of any method?

test/upgradeableProxy/DispatcherUUPS.accessControl.sol
```diff
@@ -0,0 +1,65 @@
+// SPDX-License-Identifier: UNLICENSED
```
nicopernas, line 1: If this is a test, can we rename the file to be `.t.sol`?

test/upgradeableProxy/Dispatcher.upgrade.t.sol
```diff
@@ -0,0 +1,353 @@
...
+
+        _local = LocalEnd(mars, portId, channelId, connectionHops, "1.0", "1.0");
+        _remote = CounterParty("eth2.7E5F4552091A69125d5DfCb7b8C2659029395Bdf", "channel-2", "1.0");
+
+        openChannel(_local, _remote, setting, true);
+        connectChannel(_local, _remote, setting, false, true);
+    }
+}
+
+contract DispatcherSendPacketTestSuite is ChannelOpenTestBaseSetup {
```
nicopernas, line 97: Are these tests different from the ones in Dispatcher.t.sol?

test/upgradeableProxy/Dispatcher.upgrade.t.sol
```diff
@@ -0,0 +1,353 @@
...
+        dispatcher.recvPacket(IbcReceiver(mars), IbcPacket(src, dest, 1, payload, ZERO_HEIGHT, maxTimeout), validProof);
+        vm.expectRevert(abi.encodeWithSignature("unexpectedPacketSequence()"));
+        dispatcher.recvPacket(IbcReceiver(mars), IbcPacket(src, dest, 3, payload, ZERO_HEIGHT, maxTimeout), validProof);
+    }
+
+    // TODO: add tests for unordered channel, wrong port, and invalid proof
+}
+
+// Test Chain A receives an acknowledgement packet from Chain B
+contract DispatcherAckPacketTestSuite is PacketSenderTestBase {
```
nicopernas, line 217: I'm probably overlooking something in here but where are we testing the dispatcher upgradability?

test/upgradeableProxy/Dispatcher.upgrade.t.sol
RnkSngh: the test suite is the same but the setup is different; the setup does an upgrade to a "new" DispatcherV2 implementation so this tests that there aren't any storage collisions (presumably the tests should fail on the new upgrade if there are)

test/Dispatcher.proof.t.sol
RnkSngh: I just used the `forge inspect --storage` command to see what the storage layout is and looked for the mapping from that output. Kind of manual, but probably still easier than setting up the whole handshake process?

test/upgradeableProxy/DispatcherUUPS.accessControl.sol
RnkSngh: ah yes! Will change

test/upgradeableProxy/Dispatcher.upgrade.t.sol
RnkSngh: The dispatcher is updated to a new implementation in the setup functions. [Here](https://github.com/open-ibc/vibc-core-smart-contracts/blob/8899883c42798a9b1ddfe36dffeee110712b1be1/test/upgradeableProxy/Dispatcher.upgrade.t.sol#L40-L41) is an example - these two lines are not present in the setup for the non-upgrade tests (i.e. the existing tests before this PR. Since these two lines upgrade the proxy and all tests run after this setup, the testing is done on the upgraded proxy.

test/upgradeableProxy/Dispatcher.upgrade.t.sol
RnkSngh: Though I guess the "new" Dispatcher V2 isn't much different from the old one; maybe i can change the logic in the methods a bit, or add more variables or something? Not sure if you had any ideas on specific logic we should change to test?

test/Dispatcher.t.sol
RnkSngh: [This](https://book.getfoundry.sh/forge/writing-tests#shared-setups) was the pattern I'm following; not sure exactly why they mark it as abstract; might be so that the inheriting contracts only run the setup once, or might have something to do with the test output

test/Dispatcher.proof.t.sol
nicopernas: Manual method is fine. Maybe we can comment on how to get it for future references?

test/VirtualChain.sol
```diff
@@ -34,7 +37,8 @@ struct VirtualChainData {
 
 // A test contract that keeps two types of dApps, 1. regular IBC-enabled dApp, 2. universal channel dApp
 contract VirtualChain is Test, IbcEventsEmitter {
-    Dispatcher public dispatcher;
+    IDispatcher public dispatcher;
+    Dispatcher public impl;
```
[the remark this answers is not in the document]
nicopernas: `dispatcherImplementation` and `dispatcherInterface` or `dispatcherInstance` make sense. I get that `impl` stands for implementation, my point was to be clear about the implementation of what it reffers to.

test/upgradeableProxy/Dispatcher.upgrade.t.sol
nicopernas: I guess if there were problems with the storage after the upgrade, you'd see them when calling a single of the dispatcher methods (ie `recvPacket`) right? Maybe the test can do the upgrade and call one or two methods. Easier than duplicating all the other tests. Would it make sense to test multiple upgrades too?

RnkSngh: Storage collisions can happen at any stored variable/mappings; e.g. there might be a collision in the `nextSequenceSend` mapping; so if that was the case then just testing `recvPacket` wouldn't catch it since it looks like that method only touches `nextSequence Recv`. 

My view of testing _all_ of the functionality is that it is kind of like regression testing; there is an upgrade to the code, so we should make sure all of the previous functionality is working - I guess maybe [differential testing](https://book.getfoundry.sh/forge/differential-ffi-testing) is probably the best way to do this eventually; but this might be a good substitute in the mean time. 

I also wanted to re-use this test suite to do fork testing before we deploy upgrades; just to be extra sure that the contracts in use by the prod testnet won't be negatively impacted by upgrades

The test suite runs fairly quickly (< 100ms) so I don't think there's a downside to running all of the tests on the upgrade; was your concern mainly over bloating the test suite with too much code? 

Now that I'm thinking about this again, I'm realizing maybe we should also store non-trivial state in the contract before we upgrade; to really test that that state is not impacted

test/Dispatcher.proof.t.sol
RnkSngh: added!

test/VirtualChain.sol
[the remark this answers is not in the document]
RnkSngh: done!  thanks for the suggestion!

test/upgradeableProxy/Dispatcher.upgrade.t.sol
RnkSngh: I got rid of the regression tests and  now added a small test to test for state collisions

PR #58, opened 2024-03-07
Raunak/testnet fixes

RnkSngh:
PR to address:

-  handling errors in onRecvPacket callback in try/catch
- Change the encoding to/from UniversalChannelPackets to use bytes.concat (a non-deprecated version of `abi.encodePacked` ) 

## Summary by CodeRabbit

- **New Features**
	- Enhanced error handling in packet reception through the introduction of a `try-catch` block.
	- Added virtual function behavior and error handling in contracts for improved testing of transport errors.
	- Implemented a new method for handling data extraction in the `IbcUtils` library, optimizing for performance and reliability.
- **Tests**
	- Introduced new test cases to ensure robust packet reception and to validate the integrity of data encoding and decoding processes.

### DIFF
--- a/contracts/core/Dispatcher.sol
+++ b/contracts/core/Dispatcher.sol
@@ -384,7 +384,12 @@ contract Dispatcher is IbcDispatcher, IbcEventsEmitter, Ownable, Ibc {
 
         // Not timeout yet, then do normal handling
         IbcPacket memory pkt = packet;
-        AckPacket memory ack = receiver.onRecvPacket(pkt);
+        AckPacket memory ack;
+        try receiver.onRecvPacket(pkt) returns (AckPacket memory receivedAck) {
+            ack = AckPacket(true, receivedAck.data);
+        } catch {
+            ack = AckPacket(false, ""); // Since the call to receiver.onRecvPacket failed; we have an empty ack data
+        }
         bool hasAckPacketCommitment = ackPacketCommitment[address(receiver)][packet.dest.channelId][packet.sequence];
         // check is not necessary for sync-acks
         if (hasAckPacketCommitment) {
--- a/contracts/core/Dispatcher.sol
+++ b/contracts/core/Dispatcher.sol
@@ -386,7 +386,7 @@ contract Dispatcher is IbcDispatcher, IbcEventsEmitter, Ownable, Ibc {
         IbcPacket memory pkt = packet;
         AckPacket memory ack;
         try receiver.onRecvPacket(pkt) returns (AckPacket memory receivedAck) {
-            ack = AckPacket(true, receivedAck.data);
+            ack = AckPacket(receivedAck.success, receivedAck.data);
         } catch {
             ack = AckPacket(false, ""); // Since the call to receiver.onRecvPacket failed; we have an empty ack data
         }
--- a/contracts/core/Dispatcher.sol
+++ b/contracts/core/Dispatcher.sol
@@ -387,8 +387,13 @@ contract Dispatcher is IbcDispatcher, IbcEventsEmitter, Ownable, Ibc {
         AckPacket memory ack;
         try receiver.onRecvPacket(pkt) returns (AckPacket memory receivedAck) {
             ack = AckPacket(receivedAck.success, receivedAck.data);
-        } catch {
-            ack = AckPacket(false, ""); // Since the call to receiver.onRecvPacket failed; we have an empty ack data
+        } catch Error(string memory reason) {
+            ack = AckPacket(false, bytes(reason));
+        } catch Panic(uint256 code) {
+            ack = AckPacket(false, bytes.concat("panic: ", bytes32(code)));
+        } catch (bytes memory reason) {
+            // Catch all for general revert
+            ack = AckPacket(false, reason);
         }
         bool hasAckPacketCommitment = ackPacketCommitment[address(receiver)][packet.dest.channelId][packet.sequence];
         // check is not necessary for sync-acks
--- a/contracts/examples/Mars.sol
+++ b/contracts/examples/Mars.sol
@@ -130,6 +130,7 @@ contract RevertingStringMars is Mars {
         onlyIbcDispatcher
         returns (AckPacket memory ackPacket)
     {
+        // solhint-disable-next-line
         require(false, "on recv packet is reverting");
     }
 }
@@ -158,6 +159,7 @@ contract RevertingEmptyMars is Mars {
         onlyIbcDispatcher
         returns (AckPacket memory ackPacket)
     {
+        // solhint-disable-next-line
         require(false);
     }
 }
--- a/contracts/examples/Mars.sol
+++ b/contracts/examples/Mars.sol
@@ -19,7 +19,12 @@ contract Mars is IbcReceiverBase, IbcReceiver {
 
     constructor(IbcDispatcher _dispatcher) IbcReceiverBase(_dispatcher) {}
 
-    function onRecvPacket(IbcPacket memory packet) external onlyIbcDispatcher returns (AckPacket memory ackPacket) {
+    function onRecvPacket(IbcPacket memory packet)
+        external
+        virtual
+        onlyIbcDispatcher
+        returns (AckPacket memory ackPacket)
+    {
         recvedPackets.push(packet);
 
         // solhint-disable-next-line quotes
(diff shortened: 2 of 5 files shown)


RnkSngh:
PR to address handling errors in `onRecvPacket` callback, and to fix the encoding to/from `UniversalChannelPackets`

contracts/core/Dispatcher.sol
```diff
@@ -384,7 +384,12 @@ contract Dispatcher is IbcDispatcher, IbcEventsEmitter, Ownable, Ibc {
 
         // Not timeout yet, then do normal handling
         IbcPacket memory pkt = packet;
-        AckPacket memory ack = receiver.onRecvPacket(pkt);
+        AckPacket memory ack;
+        try receiver.onRecvPacket(pkt) returns (AckPacket memory receivedAck) {
+            ack = AckPacket(receivedAck.success, receivedAck.data);
+        } catch {
+            ack = AckPacket(false, ""); // Since the call to receiver.onRecvPacket failed; we have an empty ack data
+        }
```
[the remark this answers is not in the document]
RnkSngh: @coderabbitai I don't think we need to do this since the logging should happen just outside the catch block

contracts/core/Dispatcher.sol
```diff
@@ -384,7 +384,12 @@ contract Dispatcher is IbcDispatcher, IbcEventsEmitter, Ownable, Ibc {
 
         // Not timeout yet, then do normal handling
         IbcPacket memory pkt = packet;
-        AckPacket memory ack = receiver.onRecvPacket(pkt);
+        AckPacket memory ack;
+        try receiver.onRecvPacket(pkt) returns (AckPacket memory receivedAck) {
+            ack = AckPacket(receivedAck.success, receivedAck.data);
+        } catch {
+            ack = AckPacket(false, ""); // Since the call to receiver.onRecvPacket failed; we have an empty ack data
```
nicopernas, line 391: Maybe we can extract some useful error data from the exception? The `AckPacket.data` will be interpreted as an "error message" in case `success == false`.

Maybe we can implement some of these https://docs.soliditylang.org/en/latest/control-structures.html#try-catch? what do you think?

contracts/examples/Mars.sol
```diff
@@ -111,3 +116,21 @@ contract Mars is IbcReceiverBase, IbcReceiver {
         revert UnsupportedVersion();
     }
 }
+
+/**
+ * Exact same as Mars, but reverts - used to test that transport error is seperated from errors thrown in onRecvPacket
+ */
+contract RevertingMars is Mars {
```
nicopernas, line 123: 💯

test/Dispatcher.proof.t.sol
```diff
@@ -114,6 +116,24 @@ contract DispatcherIbcWithRealProofs is IbcEventsEmitter, ProofBase {
...
+        packet.timeoutTimestamp = 15_566_401_733_896_437_760;
+        packet.dest.channelId = ch1.channelId;
+        packet.dest.portId = string(abi.encodePacked("polyibc.eth1.", IbcUtils.toHexStr(address(revertingMars))));
+        packet.src.portId = ch0.portId;
+        packet.src.channelId = ch0.channelId;
+        packet.sequence = 1;
+
+        vm.expectEmit(true, true, true, true);
+        emit WriteAckPacket(address(revertingMars), packet.dest.channelId, packet.sequence, AckPacket(false, ""));
+        dispatcher.recvPacket(revertingMars, packet, proof);
```
nicopernas, line 134: great to see this test!

nicopernas (approved):
I'm approving regardless of my previous comment. I'll leave it up to you to decide if the extra try/catch info is doable. We can always add it later too.

contracts/core/Dispatcher.sol
RnkSngh: Ah yes this is a good idea! Thanks for calling this out; this is addressed in the new commit !

contracts/libs/Ibc.sol
```diff
@@ -172,17 +172,29 @@ library IbcUtils {
...
-        internal
+    function fromUniversalPacketBytes(bytes calldata data)
+        public
         pure
         returns (UniversalPacket memory universalPacketData)
     {
-        universalPacketData = abi.decode(data, (UniversalPacket));
+        bytes32 srcPortAddr;
+        uint256 mwBitmap;
+        bytes32 destPortAddr;
```
alfredo-stonk, line 183: Is it possible to avoid allocating these fields, instead directly unpack to UniversalPacket fields?

contracts/libs/Ibc.sol
```diff
@@ -172,17 +172,29 @@ library IbcUtils {
...
 
-    function toUniversalPacketBytes(UniversalPacket memory data) internal pure returns (bytes memory packetBytes) {
-        packetBytes = abi.encode(data);
-    }
-
     // fromUniversalPacketBytes converts UniversalPacketDataBytes to UniversalPacketData, per how its packed into bytes
-    function fromUniversalPacketBytes(bytes memory data)
-        internal
+    function fromUniversalPacketBytes(bytes calldata data)
+        public
```
alfredo-stonk, line 177: Any reason from internal to public?

contracts/core/Dispatcher.sol
```diff
@@ -384,7 +384,17 @@ contract Dispatcher is IbcDispatcher, IbcEventsEmitter, Ownable, Ibc {
...
+        try receiver.onRecvPacket(pkt) returns (AckPacket memory receivedAck) {
+            ack = AckPacket(receivedAck.success, receivedAck.data);
+        } catch Error(string memory reason) {
+            ack = AckPacket(false, bytes(reason));
+        } catch Panic(uint256 code) {
+            ack = AckPacket(false, bytes.concat("panic: ", bytes32(code)));
+        } catch (bytes memory reason) {
+            // Catch all for general revert
+            ack = AckPacket(false, reason);
+        }
```
alfredo-stonk, line 397: what happens if error is none of these?

alfredo-stonk (approved):
LGTM with minor comments

contracts/libs/Ibc.sol
RnkSngh: I don't think so; I was trying this:

```
asembly {
            // Keep reusing 0x0 to move from calldata to return vars
            calldatacopy(0x0, data.offset, 32)
            universalPacketData.srcPortAddr := mload(0x0)
            calldatacopy(0x0, add(data.offset, 32), 32)
            universalPacketData.mwBitmap := mload(0x0)
            calldatacopy(0x0, add(data.offset, 64), 32)
            universalPacketData.destPortAddr := mload(0x0)
        }
```

But  the compiler doesn't like that, so that's why I went with just declaring the fields individually. Or were you thinking of some other way?

contracts/libs/Ibc.sol
RnkSngh: The `data` argument needs to be passed in as calldata to let us use `data[96:data.length]` to set the message string dynamically - solidity doesn't let you access arrays like this for non-calldata bytes. The alternative to accessing the indicies of the array would be to write a for loop in assembly which would probably be a lot more complex and not as gas optimized (also there are [vulns](https://medium.com/@JohnnyTime/solidity-smart-contract-unbounded-loops-dos-attack-vulnerability-explained-with-real-example-f4b4aca27c08) of writing for loops of unbounded length). 

Given the above implies we have to pass in `data` as calldata now, we have to make `fromUniversalPacketBytes` an external function; since solidity only passes in calldata for external functions (otherwise solidity would pass in the data as memory and that would throw a compiler error). 

Actually, I think we need to mark this as external rather than public; since I don't think there's a way to use this as an internal function from inheriting this library

contracts/libs/Ibc.sol
dshiell: Is it possible to do something like:

`universalPacketData = UniversalPacket(data[0:31], uint256(data[32:63], data[64:95], data[96:.length])`

?

contracts/libs/Ibc.sol
```diff
@@ -172,17 +172,29 @@ library IbcUtils {
...
+        assembly {
+            // Keep reusing 0x0 to move from calldata to return vars
+            calldatacopy(0x0, data.offset, 32)
+            srcPortAddr := mload(0x0)
+            calldatacopy(0x0, add(data.offset, 32), 32)
+            mwBitmap := mload(0x0)
+            calldatacopy(0x0, add(data.offset, 64), 32)
+            destPortAddr := mload(0x0)
+        }
+        universalPacketData = UniversalPacket(srcPortAddr, uint256(mwBitmap), destPortAddr, data[96:data.length]);
```
dshiell, line 193: should the final index be `data.length - 1`?

PR #60, opened 2024-03-11
ci: add job to check contract sizes

nicopernas:
## Summary by CodeRabbit

- **New Features**
    - Renamed GitHub Actions workflow to "Foundry."
    - Added a step to check contract sizes before running tests.

### DIFF
--- a/.github/workflows/foundry.yml
+++ b/.github/workflows/foundry.yml
@@ -1,4 +1,4 @@
-name: Tests
+name: Foundry
 
 on:
   pull_request:
@@ -21,5 +21,8 @@ jobs:
       - name: Install Foundry
         uses: foundry-rs/foundry-toolchain@v1
 
-      - name: Test
+      - name: Check contract sizes
+        run: forge build --sizes --deny-warnings
+
+      - name: Tests
         run: forge test


.github/workflows/foundry.yml
```diff
@@ -21,5 +21,8 @@ jobs:
       - name: Install Foundry
         uses: foundry-rs/foundry-toolchain@v1
 
-      - name: Test
+      - name: Check contract sizes
+        run: forge build --sizes
+
+      - name: Tests
```
nicopernas, line 27: I thought it would be nice to have "all foundry" commands under the same job, so we install the dependency only once. However, it also means that oversized contracts won't get the tests run... I guess that's ok but maybe someone else has strong opinions about it? @RnkSngh what do you think?

RnkSngh: Ah yes I was also meaning to ask on if we should just lump everything with the same install dependencies under one job; so i think this is a good idea. 

Does the forge build sizes actually throw on if we have contracts that are above the size limit? _Is_
 that what we want here? I think some of the test contracts might be above the size limit as well so wondering if we can somehow exclude those 
 
 
> it also means that oversized contracts won't get the tests run

Wondering if we can have the `forge test` command go before `forge build --sizes` if we're worried about this then? I think forge test should build them, and since there are no changes to contracts, forge build--sizes will just use the already compiled contracts so it should be fine to have this go after

.github/workflows/foundry.yml
```diff
@@ -21,5 +21,8 @@ jobs:
       - name: Install Foundry
         uses: foundry-rs/foundry-toolchain@v1
 
-      - name: Test
+      - name: Check contract sizes
```
RnkSngh, line 24: Might also be worth adding `forge coverage` to the ci as well. right now forge coverage is throwing a compile error, but if we add the line `src/=lib/optimism/packages/contracts-bedrock/src/` to remappings.txt file, its hould fix it (looks like it's some bug in the optimism library)

RnkSngh (approved):

.github/workflows/foundry.yml
RnkSngh: (thanks to @Inkvi  for the suggestion)

PR #63, opened 2024-03-12
add try/catch to dispatcher

RnkSngh:
PR to add try/catch to hopefully mitigate some testnet errors

There wasn't enough space remaining in the contract to add a more sophisticated try/catch that also logs the error thrown by the dapp so I settled for just logging the address of the erroring contract rather than the actual error the dapp threw . I also had to move the Ibc back into a library to reduce the contract size enough to add the simple try/catch.

I decided to leave the sophisticated try/catch around the `.onRecvPacket` callback since _some_ sophisticated error messages are better than none; and I'd imagine more of the errors are happening on that end.

 We still have some more contract space to allow for a few more of the sophisticated try/catch which log the dapp errors thrown; so let me know if there are any callbacks which would benefit from logging the actual error thrown by the dapp 

## Summary by CodeRabbit

- **New Features**
	- Introduced new error events for IBC channel operations in the blockchain contracts.
	- Added `virtual` modifier and new testing contracts with specific behaviors in the `Mars` contract.
	- Enhanced `Ibc` library with new functionalities for address conversion and channel verification.
- **Refactor**
	- Updated the `Dispatcher` contract to use interface references, improved error handling, and adjusted event emissions.
	- Refactored `Ibc` functionalities into a library.
- **Tests**
	- Modified and added new test cases for `Dispatcher` and `Ibc` functionalities, including changes in function names, logic, and variable introductions.
	- Updated test contracts to remove inheritance and utilize static calls.

### DIFF
--- a/contracts/core/Dispatcher.sol
+++ b/contracts/core/Dispatcher.sol
@@ -87,7 +87,7 @@ contract Dispatcher is IbcDispatcher, IbcEventsEmitter, Ownable {
             revert IBCErrors.invalidCounterPartyPortId();
         }
 
-        if (_isChannelOpenTry(counterparty)) {
+        if (Ibc._isChannelOpenTry(counterparty)) {
             consensusStateManager.verifyMembership(
                 proof,
                 Ibc.channelProofKey(local.portId, local.channelId),
@@ -456,7 +456,7 @@ contract Dispatcher is IbcDispatcher, IbcEventsEmitter, Ownable {
             return false;
         }
         string memory portSuffix = portId[portPrefixLen:];
-        isMatch = _hexStrToAddress(portSuffix) == addr;
+        isMatch = Ibc._hexStrToAddress(portSuffix) == addr;
     }
 
     // Prerequisite: must verify sender is authorized to send packet on the channel
@@ -504,51 +504,5 @@ contract Dispatcher is IbcDispatcher, IbcEventsEmitter, Ownable {
             // TODO: check timeoutHeight.revision_number?
             || (packet.timeoutHeight.revision_height != 0 && block.number >= packet.timeoutHeight.revision_height)
         );
-    }
-
-    /**
-     * Convert a non-0x-prefixed hex string to an address
-     * @param hexStr hex string to convert to address. Note that the hex string must not include a 0x prefix.
-     * hexStr is case-insensitive.
-     */
-    function _hexStrToAddress(string memory hexStr) internal pure returns (address addr) {
-        if (bytes(hexStr).length != 40) {
-            revert IBCErrors.invalidHexStringLength();
-        }
-
-        bytes memory strBytes = bytes(hexStr);
-        bytes memory addrBytes = new bytes(20);
-
-        for (uint256 i = 0; i < 20; i++) {
-            uint8 high = uint8(strBytes[i * 2]);
-            uint8 low = uint8(strBytes[1 + i * 2]);
-            // Convert to lowercase if the character is in uppercase
-            if (high >= 65 && high <= 90) {
-                high += 32;
-            }
-            if (low >= 65 && low <= 90) {
-                low += 32;
-            }
-            uint8 digit = (high - (high >= 97 ? 87 : 48)) * 16 + (low - (low >= 97 ? 87 : 48));
-            addrBytes[i] = bytes1(digit);
-        }
-
-        assembly {
-            addr := mload(add(addrBytes, 20))
-        }
-    }
-
-    // For XXXX => vIBC direction, SC needs to verify the proof of membership of TRY_PENDING
-    // For vIBC initiated channel, SC doesn't need to verify any proof, and these should be all empty
-    function _isChannelOpenTry(CounterParty calldata counterparty) internal pure returns (bool open) {
-        if (counterparty.channelId == bytes32(0) && bytes(counterparty.version).length == 0) {
-            open = false;
-            // ChanOpenInit with unknow conterparty
-        } else if (counterparty.channelId != bytes32(0) && bytes(counterparty.version).length != 0) {
-            // this is the ChanOpenTry; counterparty must not be zero-value
-            open = true;
-        } else {
-            revert IBCErrors.invalidCounterParty();
-        }
     }
 }
--- a/contracts/core/Dispatcher.sol
+++ b/contracts/core/Dispatcher.sol
@@ -19,7 +19,7 @@ import {
  *     Contract callers call this contract to send IBC-like msg,
  *     which can be relayed to a rollup module on the Polymerase chain
  */
-contract Dispatcher is IbcDispatcher, IbcEventsEmitter, Ownable, Ibc {
+contract Dispatcher is IbcDispatcher, IbcEventsEmitter, Ownable {
     // IBC_PortID = portPrefix + address (hex string without 0x prefix, case insensitive)
     string public portPrefix;
     uint32 public portPrefixLen;
@@ -90,23 +90,28 @@ contract Dispatcher is IbcDispatcher, IbcEventsEmitter, Ownable, Ibc {
         if (_isChannelOpenTry(counterparty)) {
             consensusStateManager.verifyMembership(
                 proof,
-                channelProofKey(local.portId, local.channelId),
-                channelProofValue(ChannelState.TRY_PENDING, ordering, local.version, connectionHops, counterparty)
+                Ibc.channelProofKey(local.portId, local.channelId),
+                Ibc.channelProofValue(ChannelState.TRY_PENDING, ordering, local.version, connectionHops, counterparty)
             );
         }
 
-        string memory selectedVersion =
-            portAddress.onOpenIbcChannel(local.version, ordering, feeEnabled, connectionHops, counterparty);
-
-        emit OpenIbcChannel(
-            address(portAddress),
-            selectedVersion,
-            ordering,
-            feeEnabled,
-            connectionHops,
-            counterparty.portId,
-            counterparty.channelId
-        );
+        string memory selectedVersion;
+        try portAddress.onOpenIbcChannel(local.version, ordering, feeEnabled, connectionHops, counterparty) returns (
+            string memory version
+        ) {
+            selectedVersion = version;
+            emit OpenIbcChannel(
+                address(portAddress),
+                selectedVersion,
+                ordering,
+                feeEnabled,
+                connectionHops,
+                counterparty.portId,
+                counterparty.channelId
+            );
+        } catch {
+            emit OpenIbcChannelError(address(portAddress));
+        }
     }
 
     /**
@@ -126,8 +131,6 @@ contract Dispatcher is IbcDispatcher, IbcEventsEmitter, Ownable, Ibc {
     ) external {
         _verifyConnectIbcChannelProof(local, connectionHops, ordering, isChanConfirm, counterparty, proof);
 
-        portAddress.onConnectIbcChannel(local.channelId, counterparty.channelId, counterparty.version);
-
         // Register port and channel mapping
         // TODO: check duplicated channel registration?
         // TODO: The call to `Channel` constructor MUST be move to `openIbcChannel` phase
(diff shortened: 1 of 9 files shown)


test/Dispatcher.proof.t.sol
```diff
@@ -48,6 +62,14 @@ contract DispatcherIbcWithRealProofs is IbcEventsEmitter, ProofBase {
...
     }
 
+    function test_ibc_channel_open_dapp_revert() public {
+        Ics23Proof memory proof = load_proof("/test/payload/channel_try_pending_proof.hex");
+
+        vm.expectEmit(true, true, true, true);
+        emit OpenIbcChannelError(address(revertingStringMars));
+        dispatcher.openIbcChannel(revertingStringMars, ch1, ChannelOrder.NONE, false, connectionHops1, ch0, proof);
+    }
+
```
nicopernas, line 72: these tests belong to `Dispatcher.t.sol` since they don't need to use proofs at all.

nicopernas: same with the other ones in the suite.

test/Dispatcher.t.sol
```diff
@@ -190,13 +193,16 @@ contract ChannelHandshakeTest is Base {
 // This Base contract provides an open channel for sub-contract tests
 contract ChannelOpenTestBase is Base {
     string portId = "eth1.7E5F4552091A69125d5DfCb7b8C2659029395Bdf";
+    string portIdI = "eth1.0xd6292A04e605AFf917Bf05b2df5dDdbdc3E35e07";
```
nicopernas, line 196:
```suggestion
    string invalidPortId = "eth1.0xd6292A04e605AFf917Bf05b2df5dDdbdc3E35e07";
```

contracts/core/Dispatcher.sol
```diff
@@ -87,26 +87,31 @@ contract Dispatcher is IbcDispatcher, IbcEventsEmitter, Ownable, Ibc {
...
-            address(portAddress),
-            selectedVersion,
-            ordering,
-            feeEnabled,
-            connectionHops,
-            counterparty.portId,
-            counterparty.channelId
-        );
+        string memory selectedVersion;
+        try portAddress.onOpenIbcChannel(local.version, ordering, feeEnabled, connectionHops, counterparty) returns (
```
nicopernas, line 99: Do we have a test the covers the case where `portAddress` is not a real contract? Will this try/catch "catch" that situation?

contracts/core/Dispatcher.sol
```diff
@@ -292,12 +303,13 @@ contract Dispatcher is IbcDispatcher, IbcEventsEmitter, Ownable, Ibc {
...
-        // delete packet commitment to avoid double timeout
-        delete sendPacketCommitment[address(receiver)][packet.src.channelId][packet.sequence];
-
-        emit Timeout(address(receiver), packet.src.channelId, packet.sequence);
+        try receiver.onTimeoutPacket(packet) {
+            // delete packet commitment to avoid double timeout
+            delete sendPacketCommitment[address(receiver)][packet.src.channelId][packet.sequence];
+            emit Timeout(address(receiver), packet.src.channelId, packet.sequence);
+        } catch {
+            emit TimeOutError(address(receiver));
```
nicopernas, line 311:
```suggestion
            emit TimeoutError(address(receiver));
```

test/Dispatcher.proof.t.sol
RnkSngh: ah yes, should be fixed now!

contracts/core/Dispatcher.sol
RnkSngh: yes thank you so much for raising this! I thought try/catch protected against EOA calls, but it actually doesn't. 

So I refactored the code to use low-level calls, and that should support all cases (e.g. reverting contracts, contracts without the handlers, and EOA) and the tests for these and it should be working now

contracts/core/Dispatcher.sol
```diff
@@ -87,26 +88,38 @@ contract Dispatcher is IbcDispatcher, IbcEventsEmitter, Ownable, Ibc {
...
         }
 
-        if (_isChannelOpenTry(counterparty)) {
+        if (Ibc._isChannelOpenTry(counterparty)) {
             consensusStateManager.verifyMembership(
                 proof,
-                channelProofKey(local.portId, local.channelId),
-                channelProofValue(ChannelState.TRY_PENDING, ordering, local.version, connectionHops, counterparty)
+                Ibc.channelProofKey(local.portId, local.channelId),
+                Ibc.channelProofValue(ChannelState.TRY_PENDING, ordering, local.version, connectionHops, counterparty)
```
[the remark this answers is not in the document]
RnkSngh: ignoring this for now since this seems like it will soon be refactored anyways in the new api changes

dshiell (approved):
lgtm

contracts/core/Dispatcher.sol
```diff
@@ -146,7 +157,20 @@ contract Dispatcher is IbcDispatcher, IbcEventsEmitter, Ownable, Ibc {
         nextSequenceRecv[address(portAddress)][local.channelId] = 1;
```
nicopernas, line 157: Since the tx will "never revert" now, do we want to do this only when the dapp callback succeeds?

contracts/core/Dispatcher.sol
```diff
@@ -160,9 +184,21 @@ contract Dispatcher is IbcDispatcher, IbcEventsEmitter, Ownable, Ibc {
             revert IBCErrors.channelNotOwnedBySender();
         }
 
-        IbcChannelReceiver reciever = IbcChannelReceiver(msg.sender);
-        reciever.onCloseIbcChannel(channelId, channel.counterpartyPortId, channel.counterpartyChannelId);
-        emit CloseIbcChannel(msg.sender, channelId);
+        IbcChannelReceiver receiver = IbcChannelReceiver(msg.sender);
+        (bool success, bytes memory data) = _try_catch(
```
nicopernas, line 188: Is `data` used?

contracts/core/Dispatcher.sol
```diff
@@ -474,6 +526,15 @@ contract Dispatcher is IbcDispatcher, IbcEventsEmitter, Ownable, Ibc {
...
     }
 
+    // Returns the result of the call if no revert, otherwise returns the error if thrown.
+    function _try_catch(address portAddress, bytes memory args) internal returns (bool success, bytes memory message) {
+        if (!Address.isContract(portAddress)) {
+            return (false, bytes("call to non-contract"));
+        }
+        // Only call if we are sure portAddress is a contract
+        (success, message) = portAddress.call(args);
+    }
```
nicopernas, line 536: this is not doing a try/catch. It's only checking if the address is a contract. I thought the whole point of the try/catch was to prevent txs being reverted due to dapps. If the dapp callback decides to revert or assert the tx will be reverted.

nicopernas (changes requested):

contracts/core/Dispatcher.sol
RnkSngh: Ah yes, I was wondering this, but I think the reason why we want to still run the logic is that we still consider it successful transport so we still want to update state;

I want to confirm with @alfredo-stonk though

contracts/core/Dispatcher.sol
RnkSngh: Data either (in the case of a success) the returned value from the low-level call, or the bytes thrown by a revert in a low level call

contracts/core/Dispatcher.sol
RnkSngh: > If the dapp callback decides to revert or assert the tx will be reverted.
If the dapp callback reverts, the low level call will just return false for success, but not revert our tx. Which is what we wanted? 
- we wanted to log an error, but not revert the tx to the dispatcher in the case of
    -  trying a call to a non-contract address
    -  trying a call to a contract which doesn't support the callback method
    -  trying a call a contract which does support the callback but which reverts the call
- we wanted to succeed all other calls and log an event (dependent on the handshake 2-step) in the case

with the current code, what happens: 
- A call to a non-contract address will return false for `success` and `call to non-contract` as the data in the `try_catch` 
- A call to a contract which doesn't support the callback interface will return false for `success` and an empty error in the `try_catch` method
- A call to a contract which does support the callback but throws in the callback will return false for `success` and the error that the contract reverted with, which will both be returned in this method
- A call to a successful contract callback will return true for success and the data that the callback returns, if any

This is what we wanted, yes? We can also verify this is how the code is working through the test suite - if you think there is another test case that is missing let me know and I can add it

RnkSngh: FYI, external calls to non-contract addresses don't revert themselves; since non-contract addresses don't have any code, they can't revert. So a low-level call will always return success for a non-contract address, which is why I have that logic in the code the way it is

RnkSngh: Our previous implementation of solidity's try/catch would have thrown errors in the case of call to a non-contract address or to a contract which doesn't have the method we are trying to call, and I don't think there is a way to account for that using solidity's try/catch because solidity's try catch only applies to a specific external call, and in these cases, the revert doesn't happen in the external call itself, but instead it happens within our dispatcher contract so the tx will still revert on the dispatcher level despite the use of solidity's try/catch 

I think the only way around this is to use low-level calls

contracts/core/Dispatcher.sol
nicopernas: `data` is assigned but not used by the caller, right?   
This will be the same? 
```solidity
        (bool success,) = _try_catch(
```

contracts/core/Dispatcher.sol
nicopernas: I can see both options being valid. it's up to us to decide. Whatever we end up doing, let's document it. Maybe a comment in the code will help. Also, we need to make it explicit to devs (regardless of what we do)

contracts/core/Dispatcher.sol
nicopernas: ah that makes sense! All cases are covered then and indeed we haven't changed the approach. Thanks for the thorough explanation.

contracts/core/Dispatcher.sol
alfredo-stonk: If we leave it this way, then dApp can send packets even when channel handshake fails, which is a big security loophole

nicopernas (approved):

contracts/core/Dispatcher.sol
RnkSngh: though not sure if we want to revert in the case of a callback failure in this step?

contracts/core/Dispatcher.sol
RnkSngh: Actually this comment made me realize we can log error data in the error events, so I pushed some changes for that

contracts/core/Dispatcher.sol
RnkSngh: Ok should be fixed now. 

I also did a similar thing for logic in the `acknowledgement` method to only do state updates in the case of success. Though for the `recvPacket` I left it as still updating the nextSequenceRecv mapping since I think we still consider successful transport in the case of a callback failure (and if it's gotten to this step then we know it's not an EOA)

RnkSngh: maybe @alfredo-stonk should verify the logic sourrounding try/catch in the `acknowledgement`, `recvPacket` and updated logic in `onConnectIbcChannel`  is indeed what we want?

RnkSngh:
> can we put all revert cases in a separate Dispatcher.revert.t.sol?
can we do the same for UCH too?

I agree this would be a fantastic organization structure, but we have a lot of pending PRs, so I wanted to wanted to first get those merged before making changes to test suites to minimize merge conflicts; so I will address reorging the test folders in another PR if that's ok?

alfredo-stonk:
> > can we put all revert cases in a separate Dispatcher.revert.t.sol?
> > can we do the same for UCH too?
> 
> I agree this would be a fantastic organization structure, but we have a lot of pending PRs, so I wanted to wanted to first get those merged before making changes to test suites to minimize merge conflicts; so I will address reorging the test folders in another PR if that's ok?

totally 👍 
also we prob don't need try-catch in UCH since Dispatcher will try-catch UCH cases too

alfredo-stonk (approved):

PR #70, opened 2024-03-27
Raunak/multiclient support

RnkSngh:
PR to add multiclient support to our dispatcher contracts! Relatively simple, though I think most of the heavy lift should be on the web2 layer.  

We needed some way of adding the mapping what connections correspond to what clientIds. Right now, this is just done through a setter; as was the previously discussed solution. The setter is permissioned; only the owner can set which connections correspond to which clients.  The corollary to this is that, since connections depend on use of a light client, using connections is now permissioned as well (i.e. no one can use a connection that hasn't had a client set for it, and this client needs to be set by the owner). If someone tries to use a connection that doesn't have a client set for it or a channel that uses a connection that doesn't have a client set for it, the tx will revert.  I'm not sure if this is indeed what we want; we might want to make the connection to clientId setter permissionless so that anyone can add what client a connection maps to. 

Also, this connection to clientId mapping is immutable - once a connection's client is set, it can't be changed, even by the owner. In the case of an unpermissioned setter for connection to client mapping, this protects malicious actors from arbitrarily setting existing connections to malicious clients.

### DIFF
--- a/contracts/core/Dispatcher.sol
+++ b/contracts/core/Dispatcher.sol
@@ -54,8 +54,7 @@ contract Dispatcher is OwnableUpgradeable, UUPSUpgradeable, IDispatcher {
     // keep track of outbound ack packets to prevent replay attack
     mapping(address => mapping(bytes32 => mapping(uint64 => bool))) private _ackPacketCommitment;
     mapping(bytes32 => string) private _channelIdToConnection;
-    mapping(string => uint256) private _connectionToClientId;
-    mapping(uint256 => LightClient) private _clientIdToLightClient;
+    mapping(string => LightClient) private _connectionToLightClient;
     uint256 private _numClients;
 
     //
@@ -88,13 +87,9 @@ contract Dispatcher is OwnableUpgradeable, UUPSUpgradeable, IDispatcher {
         OpL2StateProof calldata proof,
         uint256 height,
         uint256 appHash,
-        uint256 clientId
+        string calldata connection
     ) external returns (uint256 fraudProofEndTime, bool ended) {
-        LightClient lightClient = _clientIdToLightClient[clientId];
-        if (address(lightClient) == address(0)) {
-            revert IBCErrors.lightClientNotFound(clientId);
-        }
-        return lightClient.addOpConsensusState(l1header, proof, height, appHash);
+        return _getLightClientFromConnection(connection).addOpConsensusState(l1header, proof, height, appHash);
     }
 
     function setNewConnection(string calldata connection, LightClient lightClient) external onlyOwner {
@@ -102,8 +97,7 @@ contract Dispatcher is OwnableUpgradeable, UUPSUpgradeable, IDispatcher {
     }
 
     function removeConnection(string calldata connection) external onlyOwner {
-        uint256 clientId = _connectionToClientId[connection];
-        delete _clientIdToLightClient[clientId];
+        delete _connectionToLightClient[connection];
     }
 
     /**
@@ -581,16 +575,12 @@ contract Dispatcher is OwnableUpgradeable, UUPSUpgradeable, IDispatcher {
     }
 
     // getOptimisticConsensusState
-    function getOptimisticConsensusState(uint256 height, uint256 clientId)
+    function getOptimisticConsensusState(uint256 height, string calldata connection)
         external
         view
         returns (uint256 appHash, uint256 fraudProofEndTime, bool ended)
     {
-        LightClient lightClient = _clientIdToLightClient[clientId];
-        if (address(lightClient) == address(0)) {
-            revert IBCErrors.lightClientNotFound(clientId);
-        }
-        return lightClient.getState(height);
+        return _getLightClientFromConnection(connection).getState(height);
     }
 
     // verify an EVM address matches an IBC portId.
@@ -610,9 +600,7 @@ contract Dispatcher is OwnableUpgradeable, UUPSUpgradeable, IDispatcher {
             revert IBCErrors.invalidAddress();
         }
 
-        uint256 newClientId = ++_numClients; // Cache clientId to save SLOAD call
-        _connectionToClientId[connection] = newClientId;
-        _clientIdToLightClient[newClientId] = lightClient;
+        _connectionToLightClient[connection] = lightClient;
     }
 
     // Prerequisite: must verify sender is authorized to send packet on the channel
@@ -696,13 +684,9 @@ contract Dispatcher is OwnableUpgradeable, UUPSUpgradeable, IDispatcher {
     }
 
     function _getLightClientFromConnection(string memory connection) internal view returns (LightClient lightClient) {
-        uint256 clientId = _connectionToClientId[connection];
-        if (clientId == 0) {
-            revert IBCErrors.invalidConnection(connection);
-        }
-        lightClient = _clientIdToLightClient[clientId];
+        lightClient = _connectionToLightClient[connection];
         if (address(lightClient) == address(0)) {
-            revert IBCErrors.lightClientNotFound(clientId);
+            revert IBCErrors.lightClientNotFound(connection);
         }
     }
 }
--- a/contracts/core/Dispatcher.sol
+++ b/contracts/core/Dispatcher.sol
@@ -97,8 +97,13 @@ contract Dispatcher is OwnableUpgradeable, UUPSUpgradeable, IDispatcher {
         return lightClient.addOpConsensusState(l1header, proof, height, appHash);
     }
 
-    function addNewConnection(string calldata connection, LightClient lightClient) external onlyOwner {
-        _addNewConnection(connection, lightClient);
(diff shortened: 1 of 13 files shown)


contracts/core/Dispatcher.sol
```diff
@@ -86,11 +87,20 @@ contract Dispatcher is OwnableUpgradeable, UUPSUpgradeable, IDispatcher {
...
+        uint256 clientId
     ) external returns (uint256 fraudProofEndTime, bool ended) {
+        LightClient lightClient = _clientIdToLightClient[clientId];
+        if (address(lightClient) == address(0)) {
+            revert IBCErrors.lightClientNotFound(clientId);
+        }
         return lightClient.addOpConsensusState(l1header, proof, height, appHash);
     }
 
+    function addNewConnection(string calldata connection, LightClient lightClient) external onlyOwner {
```
alfredo-stonk, line 100: pls document the method params, esp connectionID as the first in connectionHops during channel handshake

contracts/core/Dispatcher.sol
```diff
@@ -578,6 +594,22 @@ contract Dispatcher is OwnableUpgradeable, UUPSUpgradeable, IDispatcher {
...
+        if (bytes(connection).length == 0) {
+            revert IBCErrors.invalidConnection("");
+        }
+        if (_connectionToClientId[connection] != 0) {
+            revert IBCErrors.invalidConnection(connection);
+        }
+
+        uint256 newClientId = ++_numClients; // Cache clientId to save SLOAD call
+        _connectionToClientId[connection] = newClientId;
+        _clientIdToLightClient[newClientId] = lightClient;
```
alfredo-stonk, line 610: can we directly map keccak(connectionID) -> lightClient?

contracts/core/Dispatcher.sol
```diff
@@ -86,11 +87,20 @@ contract Dispatcher is OwnableUpgradeable, UUPSUpgradeable, IDispatcher {
...
+        LightClient lightClient = _clientIdToLightClient[clientId];
+        if (address(lightClient) == address(0)) {
+            revert IBCErrors.lightClientNotFound(clientId);
+        }
         return lightClient.addOpConsensusState(l1header, proof, height, appHash);
     }
 
+    function addNewConnection(string calldata connection, LightClient lightClient) external onlyOwner {
+        _addNewConnection(connection, lightClient);
+    }
```
alfredo-stonk, line 102: can we change `addNewConnection` to `setClientForConnection` (or add as a new method) so it supports both adding new and updating existing `connectionID -> client` map?
if client == address(0), then a connection must be removed/disabled 
we might need to update existing connections' clients in the future

alfredo-stonk (commented):
still reviewing but leave these comments you can take a look first

contracts/core/Dispatcher.sol
RnkSngh: I added natspec docs in a seperate PR on [this branch](https://github.com/open-ibc/vibc-core-smart-contracts/blob/a9081b6d00c609f9409988c62384a2c16ce596e0/contracts/core/Dispatcher.sol#L842-L844), so I will implement your suggestion in that one if that's ok!

contracts/core/Dispatcher.sol
RnkSngh: Actually now that I think about it again, I think we can directly map connectionId -> lightClient, and we can get rid of the (clientID) altogether. Though I'm curious why you are suggesting to hash the conneciton id in the mapping rather than storing it directly; is that a potential gas optimisation? 

I had clientId originally as a way to "white list" what light clients can be updated in `updateClientWithOptimisticConsensusState` to avoid calling `updateClientWithOptimisticConsensusState`  on malicious contracts. Since that takes a clientId rather than a lightClient address directly, the clientId must already be included in our mapping for the tx to go through.  However, I think we can still get this same effect with passing in the connection instead, and looking up the client that maps to this connection, so we're still only calling whitelisted light clients.

contracts/core/Dispatcher.sol
RnkSngh: yes I think this is valid, should be changed now. we can have setting the client to address 0 as a seperate method

contracts/core/Dispatcher.sol
RnkSngh: added the specific comment about connectionHops [here](https://github.com/open-ibc/vibc-core-smart-contracts/blob/e732841a3900ed6c21b99b6ebbd4e5f135c5dc8d/contracts/core/Dispatcher.sol#L123-L124)

contracts/core/Dispatcher.sol
RnkSngh: tests should be added in [this commit](https://github.com/open-ibc/vibc-core-smart-contracts/pull/70/commits/062e7534ba0d92482f489cfc6deba827cbd47148)!

contracts/core/Dispatcher.sol
RnkSngh: I think having strings as the mapping should be fine; so using strings for now - should be added!

test/Dispatcher.base.t.sol
```diff
@@ -189,4 +190,28 @@ contract Base is IbcEventsEmitter, ProofBase, TestUtilsTest {
             ? abi.encodePacked('{"result":"', Base64.encode(ack.data), '"}')
             : abi.encodePacked('{"error":"', ack.data, '"}');
     }
+
+    // Store connection in channelid to connection mapping using store
+    function _storeChannelidToConnectionMapping(bytes32 channelId, bytes32 connection) internal {
```
alfredo-stonk, line 195: nit: can we pass `string connectionId` so its consistent with API param types. Can convert to bytes32 programatically in this func

contracts/core/Dispatcher.sol
```diff
@@ -86,9 +86,18 @@ contract Dispatcher is OwnableUpgradeable, UUPSUpgradeable, IDispatcher {
...
         uint256 height,
-        uint256 appHash
+        uint256 appHash,
+        string calldata connection
     ) external returns (uint256 fraudProofEndTime, bool ended) {
-        return lightClient.addOpConsensusState(l1header, proof, height, appHash);
+        return _getLightClientFromConnection(connection).addOpConsensusState(l1header, proof, height, appHash);
+    }
+
+    function setNewConnection(string calldata connection, LightClient lightClient) external onlyOwner {
```
alfredo-stonk, line 95: nit: add comment here the func can 1. set new client for a new connection; or 2. update an existing connection's to a new client.
a name like `setClientForConnection` feels more natural?

test/Dispatcher.multiclient.sol
```diff
@@ -0,0 +1,104 @@
...
+        vm.expectRevert("Ownable: caller is not the owner");
+        dispatcherProxy.removeConnection(connectionHops0[0]);
+    }
+
+    function test_addr0_channels_cannot_be_added() public {
+        vm.expectRevert(abi.encodeWithSelector(IBCErrors.invalidAddress.selector));
+        dispatcherProxy.setNewConnection(connectionHops0[0], LightClient(address(0)));
+    }
+
+    function test_Dispatcher_removeConnection() public {
```
alfredo-stonk, line 88: ada a similar test for update an existing connection so it maps to a new client?

RnkSngh: this test should be covered [here](https://github.com/open-ibc/vibc-core-smart-contracts/blob/7bcedf7e670412a8bf97e06eddeb6d65fd75dcfc/test/Dispatcher.multiclient.sol#L67-L69) no?

test/Dispatcher.base.t.sol
RnkSngh: I wanted to programmatically convert the connection from string into bytes32, but the problem is that in practice the way that solidity stores a string in memory in this array is not really reproducible (I think it might be some sort of optimization that solidity is doing) 

for example, the string "connection-0" should map to a storage slot as `0x636f6e6e656374696f6e2d300000000000000000000000000000000000000018`. When I try:

- `abi.encodePacked("connection-0")`, it produces `0x636f6e6e656374696f6e2d300000000000000000000000000000000000000000` - prefix is correct but missing the 18 at the end
-`abi.encode("connection-0")` produces  `0x636f6e6e656374696f6e2d300000000000000000000000000000000000000000` and `0x0000000000000000000000000000000000000000000000000000000000000020` which fits into two seperate storage slots.

contracts/core/Dispatcher.sol
RnkSngh: added!

contracts/core/Dispatcher.sol
RnkSngh: err now [this commit ](https://github.com/open-ibc/vibc-core-smart-contracts/pull/70/commits/77a4e5ea417b940447ab81c1bec44af1cd18dac3)after rebasing

alfredo-stonk (approved):

PR #80, opened 2024-04-10
Raunak/hex to str fix

RnkSngh:
PR to fix the IBCUtils from hex str to address conversion + adding some fuzz tests for this!


RnkSngh:
linting is failing but this should be fixed in the[ upgrade testing PR ](https://github.com/open-ibc/vibc-core-smart-contracts/pull/81)

dshiell (approved):
lgtm

contracts/libs/IbcUtils.sol
```diff
@@ -35,32 +35,35 @@ library IbcUtils {
...
+        while (i > 0) {
+            i--;
+            uint256 digit;
+            // Convert ASCII to integer value
+            if (uint8(hexBytes[i]) >= 48 && uint8(hexBytes[i]) <= 57) {
+                // '0' to '9'
+                digit = uint160(uint8(hexBytes[i]) - 48);
+            } else if (uint8(hexBytes[i]) >= 65 && uint8(hexBytes[i]) <= 70) {
+                // 'A' to 'F'
+                digit = uint160(uint8(hexBytes[i]) - 55);
```
nicopernas, line 55: please add a comment here explaining where this 55 and 87 below come from since it's not immediately obvious

RnkSngh: ah yes good idea, I added this

PR #81, opened 2024-04-10
Raunak/dispatcher upgrade testing

RnkSngh:
PR to add upgrade testing to ensure that the v2.0.0-rc3 is compatible for an upgrade with the v2.0.0-pre-audit. 

Seemed like proof generation might be a bit complex, so I'm settling for just mocking the calls to the proof verifier for now. 

One thing I did discover through this testing was that the Channel struct was indeed incompatible since we were adding a portId to it to the beginning. Moving the portId to the end of the struct should solve this issue (the change is in the Ibc.sol file). 

## Summary by CodeRabbit

- **New Features**
  - Enhanced channel and packet management in Dispatcher and UniversalChannelHandler contracts.
  - Added new functions for fraud proof and internal state management in OptimisticLightClient.
  - Introduced IbcUtils and IbcErrors libraries for improved IBC operations handling.

- **Documentation**
  - Updated README with clearer project overview and middleware introduction.
  - Added comprehensive documentation comments across various contracts and interfaces.

- **Bug Fixes**
  - Updated pragma solidity versions to ensure compatibility and security.

- **Tests**
  - Expanded test suites to cover new functionalities and scenarios in Dispatcher, including channel handshake, packet timeout, and upgrade processes.

- **Refactor**
  - Improved contract modularity and upgradeability in core contracts.
  - Refactored Proof Verifier to align with IProofVerifier interface standards.

### DIFF
--- a/test/upgradeableProxy/Dispatcher.upgrade.t.sol
+++ b/test/upgradeableProxy/Dispatcher.upgrade.t.sol
@@ -9,19 +9,35 @@ import "../../contracts/core/OptimisticLightClient.sol";
 import {ChannelHandshakeTestSuite, ChannelHandshakeTest, ChannelHandshakeUtils} from "../Dispatcher/Dispatcher.t.sol";
 import {LocalEnd} from "../utils/Dispatcher.base.t.sol";
 import {Base, ChannelHandshakeSetting} from "../utils/Dispatcher.base.t.sol";
-import {ChannelEnd, ChannelOrder, IbcEndpoint, IbcPacket, AckPacket, Ibc, Height} from "../../contracts/libs/Ibc.sol";
+import {
+    ChannelEnd,
+    ChannelOrder,
+    ChannelState,
+    IbcEndpoint,
+    IbcPacket,
+    AckPacket,
+    Ibc,
+    Height
+} from "../../contracts/libs/Ibc.sol";
 import {IbcUtils} from "../../contracts/libs/IbcUtils.sol";
-import {IbcReceiver} from "../../contracts/interfaces/IbcReceiver.sol";
+import {IbcReceiver, IbcChannelReceiver} from "../../contracts/interfaces/IbcReceiver.sol";
 import {UUPSUpgradeable} from "@openzeppelin/contracts/proxy/utils/UUPSUpgradeable.sol";
 import {OptimisticLightClient} from "../../contracts/core/OptimisticLightClient.sol";
 import {IProofVerifier} from "../../contracts/core/OptimisticProofVerifier.sol";
 import {DummyLightClient} from "../../contracts/utils/DummyLightClient.sol";
 
 import {IDispatcher} from "../../contracts/interfaces/IDispatcher.sol";
+import {DispatcherRc4} from "./upgrades/DispatcherRc4.sol";
 import {DispatcherV2Initializable} from "./upgrades/DispatcherV2Initializable.sol";
 import {DispatcherV2} from "./upgrades/DispatcherV2.sol";
+import {ERC1967Proxy} from "@openzeppelin/contracts/proxy/ERC1967/ERC1967Proxy.sol";
 
 abstract contract UpgradeTestUtils {
+    string[] connectionHops0 = ["dummy-connection-1", "dummy-connection-2"];
+    string[] connectionHops1 = ["connection-1", "connection-2"];
+    LocalEnd _localDummy;
+    ChannelEnd _remoteDummy;
+
     function upgradeDispatcher(string memory portPrefix, address dispatcherProxy)
         public
         returns (DispatcherV2Initializable newDispatcherImplementation)
@@ -31,11 +47,22 @@ abstract contract UpgradeTestUtils {
         bytes memory initData = abi.encodeWithSignature("initialize(string)", portPrefix);
         UUPSUpgradeable(dispatcherProxy).upgradeToAndCall(address(newDispatcherImplementation), initData);
     }
+
+    function deployDispatcherRC4ProxyAndImpl(string memory initPortPrefix, ILightClient initLightClient)
+        public
+        returns (IDispatcher proxy)
+    {
+        DispatcherRc4 dispatcherImplementation = new DispatcherRc4();
+        bytes memory initData =
+            abi.encodeWithSelector(DispatcherRc4.initialize.selector, initPortPrefix, initLightClient);
+        proxy = IDispatcher(address(new ERC1967Proxy(address(dispatcherImplementation), initData)));
+    }
 }
 
 contract ChannelHandShakeUpgradeUtil is ChannelHandshakeUtils {
     uint32 nextSequenceSendSlot = 153;
     uint32 sendPacketCommitmentSlot = 156;
+    uint32 ackPacketCommitmentSlot = 158;
     uint32 nextSequenceAckSlot = 155;
     uint32 nextSequenceRecvSlot = 154;
     IbcPacket[3] packets;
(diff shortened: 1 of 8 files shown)


test/upgradeableProxy/Dispatcher.upgrade.t.sol
```diff
@@ -36,43 +45,67 @@ abstract contract UpgradeTestUtils {
...
                 re.version = versions[j];
+                remote.versionCall = versions[j];
+                remote.versionExpected = versions[j];
+                local.version = versions[j];
                 channelOpenInit(le, re, settings[i], true);
-                channelOpenTry(le, re, settings[i], true);
+                channelOpenTry(remote, local, settings[i], true);
                 channelOpenAck(le, re, settings[i], true);
-                channelOpenConfirm(le, re, settings[i], true);
+                channelOpenConfirm(remote, local, settings[i], true);
```
nicopernas, line 79: why do we need `le`,`re`, `local` and `remote`? wouldn't `le`,`re` be enough?

test/upgradeableProxy/Dispatcher.upgrade.t.sol
```diff
@@ -36,43 +45,67 @@ abstract contract UpgradeTestUtils {
...
+        ChannelHandshakeSetting memory setting = ChannelHandshakeSetting(ChannelOrder.ORDERED, false, true, validProof);
+
+        LocalEnd memory remote =
+            LocalEnd(receivingDapp, re.portId, re.channelId, le.connectionHops, le.versionCall, le.versionExpected);
+
+        ChannelEnd memory local = ChannelEnd(le.portId, le.channelId, le.versionCall);
+        channelOpenInit(le, re, setting, true);
+        channelOpenTry(remote, local, setting, true);
+        channelOpenAck(le, re, setting, true);
+        channelOpenConfirm(remote, local, setting, true);
```
nicopernas, line 96: same here

test/upgradeableProxy/Dispatcher.upgrade.t.sol
```diff
@@ -106,41 +149,70 @@ contract ChannelHandShakeUpgradeUtil is ChannelHandshakeUtils {
         bytes32 slot1 = keccak256(abi.encode(portAddress, nextSequenceRecvSlot));
         slot = keccak256(abi.encode(channelId, slot1));
     }
+
+    function mockHandShakeProofs(LocalEnd memory _local, ChannelEnd memory _remote) public {
+        // Mock the proofs required to initialize a channel
+        mockProofs(ChannelState.TRY_PENDING, _remote.portId, _local.portId, _remote.channelId, _local.channelId);
+        mockProofs(ChannelState.ACK_PENDING, _local.portId, _remote.portId, _local.channelId, _remote.channelId);
+        mockProofs(ChannelState.CONFIRM_PENDING, _remote.portId, _local.portId, _remote.channelId, _local.channelId);
+    }
```
nicopernas, line 158: why do you need to mock proofs? can't you use a dummy verifier in this case? the proof verifier holds no state so it shouldn't matter too much in the upgrade path right?

test/upgradeableProxy/Dispatcher.upgrade.t.sol
RnkSngh: the 4 channel helpers here take a LocalEnd as the first type and a ChannelEnd as the second type. Because we want to flip-flop passing in the local and remote channel ends to these 4 handshake helper methods, we need the `local` and `remote` to store the other analagous types.

- `le` is the local end as a `LocalEnd` type
- `local` is the local end as a `ChannelEnd` type
- `re` is the remote end as a `ChannelEnd` type
- `remote` is the remote end as a `LocalEnd` type

test/upgradeableProxy/Dispatcher.upgrade.t.sol
RnkSngh: should be updated to only use dummy client now

contracts/core/UniversalChannelHandler.sol
```diff
@@ -45,6 +45,15 @@ contract UniversalChannelHandler is IbcReceiverBaseUpgradeable, UUPSUpgradeable,
...
     }
 
+    /**
+     * @dev Close a universal channel.
+     * Cannot send or receive packets after the channel is closed.
+     * @param channelId The channel id of the channel to be closed.
+     */
+    function closeChannel(bytes32 channelId) external onlyOwner {
+        dispatcher.channelCloseInit(channelId);
+    }
```
nicopernas, line 55: this should be part of the previous PR (ch close)

RnkSngh: ah yes - good catch. I added this change to[ this commit ](https://github.com/open-ibc/vibc-core-smart-contracts/pull/100/commits/53cb9230c2bd9fdb82f0cf8cebaa3a2967372690)on the v2 branch, so it should be included there instead now

test/upgradeableProxy/Dispatcher.upgrade.t.sol
```diff
@@ -31,48 +47,81 @@ abstract contract UpgradeTestUtils {
...
-                ChannelEnd memory re = _remote;
                 le.versionCall = versions[j];
                 le.versionExpected = versions[j];
                 re.version = versions[j];
+                remote.versionCall = versions[j];
+                remote.versionExpected = versions[j];
+                local.version = versions[j];
                 channelOpenInit(le, re, settings[i], true);
-                channelOpenTry(le, re, settings[i], true);
+                channelOpenTry(remote, local, settings[i], true);
```
dshiell, line 93: did remote and local get reversed?

nicopernas: +1

test/upgradeableProxy/Dispatcher.upgrade.t.sol
nicopernas: consider renaming them to 
- `le` -> `localEnd`
- `local` -> `localChannelEnd`
- `re` -> `remoteChannelEnd`
- `remote` -> `remoteEnd`

test/upgradeableProxy/Dispatcher.upgrade.t.sol
nicopernas: please consider renaming these for clarity

test/upgradeableProxy/Dispatcher.upgrade.t.sol
RnkSngh: Shouldn't they be reversed on each call, since they will be on separate sides (chanOpen init and ack happens on src chain, while chanOpen try and confirm happens on dest)? We do a similar thing of reversing them [here](https://github.com/open-ibc/vibc-core-smart-contracts/blob/25e9a3d911ee8c45218021aee2c230abb03f335b/test/VirtualChain.sol#L139-L141), so that's what I was following

alfredo-stonk: I think people including me get confused seeing the code changes. If this is the right way, then how previous "wrong" params pass the tests? If both works, then we have a bigger problem to fix

[Previous implementation](https://github.com/open-ibc/vibc-core-smart-contracts/blob/25e9a3d911ee8c45218021aee2c230abb03f335b/test/VirtualChain.sol#L139-L141) uses symmetrical params and avoid dup code. 
Tests here seems convoluted:
- `LocalEnd` already has a `IbcChannelReceiver` field.  
- Why local and remote are not the same type?`LocalEnd memory le, ChannelEnd memory re` 
- Why do multi channel handshakes in one call?
- channel version is part of `ChannelSetting`

test/upgradeableProxy/upgrades/DispatcherRc4.sol
```diff
@@ -0,0 +1,6352 @@
...
+     * @dev verifies the provided ICS23 proof given the trusted app hash. Reverts in case of failure.
+     *
+     * @param appHash trusted l2 app hash (state root)
+     * @param key key to be proven non-existing
+     * @param proof ICS23 non-membership proof
+     */
+    function verifyNonMembership(bytes32 appHash, bytes calldata key, Ics23Proof calldata proof) external pure;
+}
+
+// lib/base64/base64.sol
```
alfredo-stonk, line 93: Any reason lib code is included here instead of lib imports (to avoid adding 1k+ LOC)?

RnkSngh: This is to separate out the V2.0.0-RC4 version of the code with what is currently present in the repo, since some of the dependencies of the Dispatcher contract also are different, and I think this was the easiest way of maintaining two different copies.

see also [this comment ](https://github.com/open-ibc/vibc-core-smart-contracts/pull/96#discussion_r1591810796)

RnkSngh: so we can't import from lib because that contains the current version of libs, not the v2.0.0-RC4 version

test/upgradeableProxy/Dispatcher.upgrade.t.sol
RnkSngh: should be renamed now for clarity

test/upgradeableProxy/Dispatcher.upgrade.t.sol
RnkSngh: should be renamed now

test/upgradeableProxy/Dispatcher.upgrade.t.sol
RnkSngh: My original intention was to change these tests to better model the local and remote ends as seperate chains, but I see how this is very confusing  (and perhaps an incorrect test) given I'm mixing this implementation with our unitests which don't use the virtual chain.

I think the resolution here is to refactor this suite to use virtual chain, I will push these changes shortly

> If this is the right way, then how previous "wrong" params pass the tests?
The previous version assumed the local and remote ends were both on the same chain. This new version assumes that they are on different chains. It makes sense both of these two tests pass because the only access control is around proof verification, and we're using dummy clients for this test

test/upgradeableProxy/upgrades/DispatcherRc4.sol
alfredo-stonk: You can [forge install](https://book.getfoundry.sh/reference/forge/forge-install) one dep with multi versions, eg.
```sh
# install specific version with remapping
forge install openzeppelin-contracts-solc-0.7=openzeppelin/openzeppelin-contracts@v3.4.2-solc-0.7

# install the latest version
forge install openzeppelin/openzeppelin-contracts
```
Our repo `open-ibc/vibc-core-smart-contracts` can be treated as a dep too if needed

dshiell (approved):
lgtm, approving to unblock

test/upgradeableProxy/upgrades/DispatcherRc4.sol
RnkSngh: Ah I see, you meant install it as a forge lib. I chose not to do a forge install and instead just flatten this file because the compile times are already pretty large (> 20 seconds) - and adding this as a lib seems overkill? I can see your point though as it is a bit cleaner, but practically it might just be easier to have this be flattened file?

RnkSngh: I can go either way , if you have strong opinions i'm happy to add as an external foundry lib

test/upgradeableProxy/Dispatcher.upgrade.t.sol
RnkSngh: acutally, I think re-facorting to use Virtual chain will be slightly more complex than I originally thought. Seeing as we have a few different PRs we still need to merge, I will save this as a later task once we get everything we need sorted

RnkSngh: so for now, I think we can just stick to not switching local and remote for now

test/upgradeableProxy/upgrades/DispatcherRc4.sol
alfredo-stonk: I don't understand how importing lib code increase compile time compared to flattened file. Can you explain?

RnkSngh: when running `forge compile`, it would compile _everything_ in the submodule (including the sub-modules for the vibc-core repo, files not related to UCH , test files, etc) whereas the flattened file should only contain the files relevant to the UCH contract. For whatever reason, it seems to be the optimism submodule in the vibc core repo that is adding a lot of compile time; UCH doesn't use this so these aren't imported in the flattened file ; so I think this alone could lead to a reduction in compile time. 

I haven't actually tested this though - might be different in practice

RnkSngh: resolving this for now to merge, added as an isuse on the kanban to resolve later a

test/upgradeableProxy/Dispatcher.upgrade.t.sol
RnkSngh: resolving this for now to merge, added as an isuse on the kanban to resolve later a

alfredo-stonk: Yea, previous params were wrong then or rather testing undefined behavior. Channel handshake 4 steps should be alternating between A and B.

PR #83, opened 2024-04-11
fix writeTimeoutPacket vuln

RnkSngh:
Quick PR to fix the vuln in writetimeoutpacket to check for proofs before writing a timeout. Since we're checking for a proof of the packet existing in the ibc state, an arbitrary packet with only the timeout changed cannot be spoofed now. 

Note that this doesn't patch the vuln for channels which use the dummy light client, though this seems to be expected that these channels in general are not secure

## Summary by CodeRabbit

- **New Features**
  - Enhanced security for packet processing in the Dispatcher contract by requiring an additional verification proof (`Ics23Proof`).
  
- **Tests**
  - Added a new test to validate the prevention of packet spoofing in timeout scenarios.

- **Refactor**
  - Updated the `writeTimeoutPacket` function across contracts and interfaces to include proof verification.

### DIFF
--- a/contracts/core/Dispatcher.sol
+++ b/contracts/core/Dispatcher.sol
@@ -454,23 +454,6 @@ contract Dispatcher is OwnableUpgradeable, UUPSUpgradeable, IDispatcher {
         emit WriteAckPacket(receiver, packet.dest.channelId, packet.sequence, ack);
     }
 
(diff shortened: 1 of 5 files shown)


dshiell (approved):
this looks correct to me!

alfredo-stonk (approved):

PR #96, opened 2024-04-25
Raunak/uch upgrade testing

RnkSngh:
PR to test upgrades to our uch proxy 

The main state we need to worry about in UCH is the `mwStackAddrs` array so this testing orients around that, and tests both a direct Earth-> UCH connection and also one with an extra middleware layer(i.e. Earth -> GeneralMiddleware -> UCH) .

## Summary by CodeRabbit

- **New Features**
	- Enhanced upgrade capabilities for Universal Channel Handler, supporting transitions up to version 3.
- **Tests**
	- Expanded testing suite with new imports and functions to support comprehensive testing of upgrade processes.

### DIFF
--- a/test/upgradeableProxy/DispatcherRC4.upgrade.t.sol
+++ b/test/upgradeableProxy/DispatcherRC4.upgrade.t.sol
@@ -5,6 +5,7 @@ import "forge-std/console2.sol";
 import {DispatcherUpdateClientTestSuite} from "../Dispatcher/Dispatcher.client.t.sol";
 import {DispatcherIbcWithRealProofsSuite} from "../Dispatcher/Dispatcher.proof.t.sol";
 import {Mars} from "../../contracts/examples/Mars.sol";
+import {Earth} from "../../contracts/examples/Earth.sol";
 import "../../contracts/core/OptimisticLightClient.sol";
 import {ChannelHandshakeTestSuite, ChannelHandshakeTest, ChannelHandshakeUtils} from "../Dispatcher/Dispatcher.t.sol";
 import {LocalEnd} from "../utils/Dispatcher.base.t.sol";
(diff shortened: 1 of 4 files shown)


test/upgradeableProxy/DispatcherRC4.upgrade.t.sol
```diff
@@ -152,21 +196,35 @@ contract DispatcherRC4MidwayUpgradeTest is ChannelHandShakeUpgradeUtil, UpgradeT
         dummyLightClient2 = new DummyLightClient();
 
         mars = new Mars(dispatcherProxy);
+        uch = deployUCHV2ProxyAndImpl(address(dispatcherProxy));
+        earth = new Earth(address(uch));
+        // Register mw stack and test it isn't corrupted after the upgrade
+        mw1 = new GeneralMiddleware(1 << 1, address(uch));
+        mwBitmap = uch.MW_ID() | mw1.MW_ID();
```
dshiell, line 203: I think UCH is actually a dapp itself instead of middleware. At least this is what I recall from the conversation at the last offsite.

test/upgradeableProxy/upgrades/UCHV2.sol
dshiell, line 1: Why is this file so big? If I understand its just a v2 of the UCH contract to make sure we can upgrade?

test/upgradeableProxy/DispatcherRC4.upgrade.t.sol
RnkSngh: This test is to ensure that the `mwStackAddrs` stored in the uch aren't impacted by upgrades (to test for storage collisions). Are you suggesting that we change our UCH to get rid of the `mwStackAddrs` mapping (and all other mw related methods in the uch) altogether to get rid of UCH's middleware functionality? I'm not sure how much of what we discussed on the onsite is considered finalized design?

dshiell (approved):
Lgtm

PR #98, opened 2024-04-30
add npm package

RnkSngh:
PR to add an npm package which supports
- contract deployments and upgrades via importing javascript code
- contract deployments and upgrades via command line 

see README.md in root for details! 

## Summary by CodeRabbit

- **New Features**
  - Introduced enhanced Ethereum account management, allowing users to load accounts from various sources such as private keys, mnemonics, or keystore directories.
  - Added functionalities for verifying proofs in Ethereum contracts, including membership, non-membership, and state updates.
  - Implemented support for Inter-Blockchain Communication (IBC) with new methods for packet handling and event management.
  - Introduced a dummy light client and an optimistic light client for Ethereum, providing methods for state verification and fraud proof handling.

- **Enhancements**
  - Added new interfaces and types for improved contract interactions and event handling.
  - Expanded methods for managing Ethereum contracts, including ownership transfer and event filtering.

- **Bug Fixes**
  - Improved data structures and schemas for more accurate contract configuration and validation.


dshiell (approved):
approved, but talk to @nicopernas

hardhat.config.cjs
```diff
@@ -0,0 +1,6 @@
+require("@nomicfoundation/hardhat-toolbox");
+
+/** @type import('hardhat/config').HardhatUserConfig */
+module.exports = {
+  solidity: "0.8.24",
```
nicopernas, line 5: should we add the same version to foundry.toml?

package.json
```diff
@@ -1,13 +1,82 @@
 {
-  "name": "vibc-core-smart-contracts",
-  "version": "1.0.0",
-  "main": "index.js",
-  "repository": "https://github.com/open-ibc/vibc-core-smart-contracts",
+  "name": "@open-ibc/vibc-core-smart-contracts",
+  "version": "0.1.3",
```
nicopernas, line 3: this should match with the current version.

package.json
```diff
@@ -1,13 +1,82 @@
...
+    "hardhat-gas-reporter": "^1.0.8",
+    "solidity-coverage": "^0.8.0",
+    "tsup": "^8.0.2"
   },
   "scripts": {
-    "lint": "solhint contracts/**/*.sol"
+    "lint": "solhint contracts/**/*.sol",
+    "test": "forge test",
+    "build": "npm run gen-types && tsup",
+    "gen-types": "npm run build-contracts && typechain --target ethers-v6 --out-dir src/evm/contracts/ './out/?(OptimisticProofVerifier|ProofVerifier|Ibc|IbcUtils|Channel|Dispatcher|Mars|Earth|UniversalChannelHandler|DummyProofVerifier|DummyLightClient|ERC1967Proxy|OptimisticLightClient).sol/*.json'",
```
nicopernas, line 43: can this be `generate-ts-bindings` so we can then have another one saying `generate-golang-bindings`?

package.json
```diff
@@ -1,13 +1,82 @@
...
+    "build-contracts": "forge build",
+    "deploy-contracts": "npm run build && node dist/deploy.js",
+    "deploy-simple": "node dist/deploy.js",
+    "prepublish": "npm run build"
```
nicopernas, line 47: can we also make sure we build the bindings before we publish? Maybe this is not the right place for that but it would be nice to add some checks that will prevent us from "forgetting about updating the bindings" with each release.

specs/contracts.spec.yaml
```diff
@@ -0,0 +1,77 @@
...
+    - name: 'contracts/libs/Ibc.sol:Ibc'
+      address: '{{Ibc}}'
+  deployer: 'KEY_POLYMER'
+
+- name: DispatcherProxy
+  description: 'Dispatcher proxy contract'
+  factoryName: 'ERC1967Proxy'
+  deployArgs:
+    - '{{Dispatcher}}'
+    - '$INITARGS'
```
nicopernas, line 37: This comment is on me really since I didn't add the necessary docs about `$INITARGS` but can we add a line or two explaining what this actually does?

src/scripts/deploy-script.ts
```diff
@@ -0,0 +1,53 @@
...
+  parseObjFromFile,
+} from "..";
+
+import { getMainLogger } from "../utils/cli";
+import { DEFAULT_RPC_URL } from "../utils/constants";
+
+// can either parse
+const accountConfigFromYaml = {
+  name: "local",
+  registry: parseObjFromFile("specs/evm.accounts.yaml"),
```
nicopernas, line 16: this works while you are executing the script from the right directory, which I guess is the expected situation when running `npm run deploy`. However, this makes me wonder how will this work when used from `ibctl` for example? Let's keep that in mind anyway, it would be nice not having to duplicate the yaml file :)

src/scripts/deploy-script.ts
```diff
@@ -0,0 +1,53 @@
...
+const accounts = AccountRegistry.loadMultiple([accountConfig]);
+const contracts = ContractRegistryLoader.loadSingle(
+  parseObjFromFile("specs/contracts.spec.yaml")
+);
+
+const chain: Chain = {
+  rpc: process.env.RPC_URL ? process.env.RPC_URL : DEFAULT_RPC_URL,
+  chainId: process.env.DEPLOYMENT_CHAIN_ID
+    ? parseInt(process.env.DEPLOYMENT_CHAIN_ID)
+    : 31337,
```
nicopernas, line 41: what does this number mean?

src/utils/constants.ts
```diff
@@ -0,0 +1,11 @@
+export const DEFAULT_DEPLOYER = "default";
+export const DEFAULT_RPC_URL = "http://127.0.0.1:8545";
+export const DEFAULT_CHAIN_ID = 31337;
```
nicopernas, line 3: ok. it's defined here as a constant. maybe use this instead of the magic number?

tsconfig.json
```diff
@@ -0,0 +1,112 @@
+{
+  "compilerOptions": {
```
nicopernas, line 2: kudos for listing all the available options!!

hardhat.config.cjs
RnkSngh: actually thanks for leaving this comment - I think we can get rid of hardhat altogether, I don't think it makes sense to have it in this package; hardhat would be more useful for users of this npm package

package.json
RnkSngh: By current version, I'm assuming you mean the version that the main branch of this repo is tagged with right? 

if so, I'm leaving the version to a lower version for now so that we can set it to the current version once we finalize this PR (since there might be a couple more iterations before we finalize)

package.json
RnkSngh: Yes I can rename it; though `generate-golang-bindings` would be run by `abigen` right? So Since we have to run `gen-types` with npm, I actually don't think we would have a naming collision?

package.json
RnkSngh: not sure I understand your comment, `npm run build` calls `npm run gen-types && tsup`, and `npm run gen-types` generates the bindings. Since we've included `npm run build` in our prepublish, it should always run this command before publishing? So I don't think we need to check for updating bindings, since they should automatically be done on publish

src/scripts/deploy-script.ts
RnkSngh: this is the default chain ID used by hardhat (and maybe anvil)

src/utils/constants.ts
RnkSngh: ahh good catch, will update - thanks!

package.json
RnkSngh: npm is pretty strict about allowing to re-publish versions so I think we should only publish to v2.1.0 once we're sure it's finalized

src/scripts/deploy-script.ts
RnkSngh: ah good point - i think we can add env variables to configure - I added them here: https://github.com/open-ibc/vibc-core-smart-contracts/blob/2ff5078d4b3cd165d0ecc4cfe3f22e12319abb2b/src/utils/constants.ts#L27-L37

so for ibctl (or other use cases, we should just be able to specify the `SPECS_BASE_PATH` env variable and that will override the default files

specs/contracts.spec.yaml
RnkSngh: should be done now! and I added comments for all arguments in the yaml files

src/utils/constants.ts
RnkSngh: should be updated now!

package.json
nicopernas: When this PR is merged we should bump the v2.1.0 to v2.2 or v3. Let's keep the one version for tags and "node version" to keep things simple

package.json
nicopernas: It shouldn't matter what command you use, right? these names are the "npm script names". You should end up with something like this 
```json
    "generate-ts-bindings": "npm run build-contracts && typechain --target ethers-v6 --out-di...",
    "generate-golang-bindings": "npm run build-contracts && abigen --abi ...",
```

package.json
nicopernas: I didn't realise that `npm run build` called the binding construction target as well. Maybe it should be renamed to `build-contracts-and-bindings` ?

src/scripts/deploy-script.ts
nicopernas: hmm can we bundle the contract yaml with the package instead? I would expect not having to set up any of this when running `ibctl`. You should be able to run this 

```sh
$ ibctl deploy ... # uses the bundled "default" yaml

$ ibctl deploy --contracts '/path/to/my-spec.yaml' # in case I want to override the default one. 
```

src/scripts/deploy-script.ts
nicopernas: 🙏

package.json
RnkSngh: npm run build also runs `tsup`, which also bundles _all_ typescript files into javascript in a releasable version (which is also another step we need to do before publishing). So since it also bundles the whole package, I don't think calling it `build-contracts-and-bindings` is accurate? I think it's convention with npm packages to have the generic "build" command build _everything_, including the typescript files _and_ any bindings/types/modules etc

 We already do have a `gen-types` command which might be what you are thinking of, I think we can rename that to `build-contracts-and-bindings` in case that would be clearer naming

src/scripts/deploy-script.ts
RnkSngh: ah yes adding cli options might be a better idea than using env variables. 

The first command should already work but I can work on adding the path as a cli arg instead of an env variable

package.json
RnkSngh: renamed the gen-types to `build-ts-contract-bindings`! Though let me know if this can still be made clearer!

package.json
RnkSngh: ah gotcha - I was initially confused since I wasn't sure why we would need an npm command to generate go bindings; but this seems to be a general practice. So for now I can rename just in case. 

Renamed gen-types to build-ts-contract-bindings and added a dummy build-go-contract-bindings for now! Thanks for the feedback!

src/scripts/deploy-script.ts
RnkSngh: added CLI parsing too now! So the order in which we search for arguments, in order of priority: 

- first search for any args passed into CLI
- then search for ENV variables
- then use default values

nicopernas (approved):

PR #104, opened 2024-05-07
remove mw functionaity from uch

RnkSngh:
PR to remove mw functionality from the UCH to simply uch logic. this should be backwards compatible the uch currently deployed on testnet and should allow us to do a UUPS upgrade 

## Summary by CodeRabbit

- **New Features**
	- Added `IbcMwPacketSender` interface to improve middleware communication capabilities.

- **Refactor**
	- Simplified the middleware handling by removing specific middleware stack addresses and functions related to middleware packet sending in the `UniversalChannelHandler`.
	- Removed inheritance of `IbcMwPacketSender` from `IbcMiddlewareProvider` to streamline interface relationships.

- **Tests**
	- Adjusted tests to skip certain middleware sender scenarios reflecting the current support limitations in middleware handling.

- **Bug Fixes**
	- Updated logic in packet reception and middleware handling to enhance reliability and performance.

### DIFF
--- a/test/upgradeableProxy/DispatcherRC4.upgrade.t.sol
+++ b/test/upgradeableProxy/DispatcherRC4.upgrade.t.sol
@@ -172,19 +172,6 @@ contract DispatcherRC4MidwayUpgradeTest is ChannelHandShakeUpgradeUtil, UpgradeT
     IUniversalChannelHandler uch;
     LocalEnd _localUch;
     Earth earth;
-    uint256 mwBitmap;
-
-    GeneralMiddleware mw1;
-
-    event RecvMWPacket(
-        bytes32 indexed channelId,
-        bytes32 indexed srcPortAddr,
-        bytes32 indexed destPortAddr,
-        // middleware UID
-        uint256 mwId,
-        bytes appData,
-        bytes mwData
-    );
 
     function setUp() public override {
         ChannelHandshakeSetting memory setting = ChannelHandshakeSetting(ChannelOrder.ORDERED, false, true, validProof);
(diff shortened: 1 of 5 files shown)


test/upgradeableProxy/DispatcherRC4.upgrade.t.sol
```diff
@@ -279,8 +260,8 @@ contract DispatcherRC4MidwayUpgradeTest is ChannelHandShakeUpgradeUtil, UpgradeT
         assertEq(2, nextSequenceSendValue);
 
         // Now recv uch packet
-
         bytes memory appData = abi.encodePacked("hello using mw stack");
+        uint256 mwBitmap = 1;
```
dshiell, line 264: what does "1" mean here? is it mw1.MW_ID()?

dshiell (commented):
one small question otherwise lgtm

test/upgradeableProxy/DispatcherRC4.upgrade.t.sol
RnkSngh: yes, it should be the MW_ID (since there's only one MW in the stack)

I'm not sure if we should get rid of the MW_ID stored in the UCH (and also if we should remove how it is currently encoded in the universal packet data [here](https://github.com/open-ibc/vibc-core-smart-contracts/blob/d804d0d47cb4b1c4af353b959dbdc7baa9d48a04/contracts/core/UniversalChannelHandler.sol#L97)) - it seems like we don't need it anymore, but it might break dapps who are expecting this MW_ID to be encoded in the previous format with MW_ID included?

dshiell (approved):
approving to unblock

test/upgradeableProxy/DispatcherRC4.upgrade.t.sol
dshiell: Should we use "mw1.MW_ID()" here instead of "1" then? This is just a nit so I've approved the PR.

RnkSngh: sure, changed it to be `uch.MW_ID`

PR #105, opened 2024-05-07
add channelid and port information in dispatcher proxy error events

RnkSngh:
PR to respond to the issue to make errors emitted in the dispatcher contain information like channeld and portId that can be used to track which packet/channel caused the error. 

Previously, we couldn't implement this because the contract size was too large, and adding events increased it over the size limit. But now since we have gotten rid of the timeout implementation, we have some more size to work with, and can include these changes in mainnet launch to make things easier on the indexer, if we want. 

## Summary by CodeRabbit

- **New Features**
  - Enhanced event logging in smart contracts to include additional parameters such as `counterpartyPortId`, `localPortId`, `localChannelId`, and `channelId` for better context in channel operations.

- **Refactor**
  - Updated event emissions and error message encodings across various contracts to reflect new parameter inclusions, improving traceability and debugging capabilities.

- **Tests**
  - Modified test cases to align with the updated event parameter requirements, ensuring consistency and reliability in contract behavior validations.

### DIFF
--- a/contracts/core/Dispatcher.sol
+++ b/contracts/core/Dispatcher.sol
@@ -181,7 +181,7 @@ contract Dispatcher is OwnableUpgradeable, UUPSUpgradeable, ReentrancyGuard, IDi
                 msg.sender, abi.decode(data, (string)), ordering, feeEnabled, connectionHops, counterpartyPortId
             );
         } else {
-            emit ChannelOpenInitError(msg.sender, data);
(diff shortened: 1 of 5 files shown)


dshiell:
Is this a temporary contract size decrease which will run into issues once we add back timeout implementation?

RnkSngh:
Closing PR for now to avoid concerns with having to remove events to deal with contract size if timeout implementation ends up being too big

PR #108, opened 2024-05-29
Raunak/audit fixes 05 2024

RnkSngh:
PR to fix most issues identified in Bailsec's most recent audit! 

Specifically, this PR is to fix:

- 20: Unused imports
- 04: Function selector mismatch of onChanOpenConfirm
- 09: timeoutTimestamp can be in the past
- 28: IbcReceiverUpgradeable is not compatible with proxies
- 30:The _hexStrToAddress function doesn’t check if the string is a valid hex string
- 32: Unsafe downcasting could lead to unexpected behaviour
- 12: sendPacket lacks nonReentrant modifier
- 03: Blunder within writeTimeoutPacked function allows for marking received packet as timed out
- 21: Comment indicates version setting during _connectChannel is incorrect

And to clarify in comments (but not fix):
-  10: Unequal block.numbers amongst chains may result in unexpected timeout
- 25: Any counterparty can open a channel with UniversalChannelHandler
- 06: Blunder in channel closure mechanism can result in unidirectional counter channel still staying open

each commit message in this PR contains the issue number which it addresses

## Summary by CodeRabbit

- **New Features**
  - Added channel authorization functionality in `DispatcherRC4MidwayUpgradeTest`.

- **Bug Fixes**
  - Improved packet handling and validation in `Dispatcher.sol`.
  - Enhanced error handling by adding new error definitions in `IBCErrors.sol`.

- **Refactor**
  - Simplified and optimized import statements across multiple contracts.
  - Restructured inheritance and authorization logic in `Earth.sol`.

- **Tests**
  - Added new test functions for packet validation and timeout scenarios.
  - Updated test files to reflect new import structures and added error handling checks.

### DIFF
--- a/contracts/core/Dispatcher.sol
+++ b/contracts/core/Dispatcher.sol
@@ -148,7 +148,8 @@ contract Dispatcher is OwnableUpgradeable, UUPSUpgradeable, ReentrancyGuard, IDi
      * @notice Initializes the channel opening process with the specified parameters. This is the first step in the  channel
      * handshake, initiated directly by the dapp which wishes to establish a channel with the receiver.
      * @param ordering The ordering of the channel (ORDERED or UNORDERED).
-     * @param feeEnabled A boolean indicating whether fees are enabled for the channel.
+     * @param feeEnabled A boolean indicating whether fees are enabled for the channel. Note: This value isn't currently
+     * used
      * @param connectionHops The list of connection hops associated with the channel, with the first channel in this
      * array always starting from the chain this contract is deployed on
      * @param counterpartyPortId The port ID of the counterparty.
@@ -193,7 +194,7 @@ contract Dispatcher is OwnableUpgradeable, UUPSUpgradeable, ReentrancyGuard, IDi
      * chain.
      * @param local The counterparty information for the receiver.
      * @param ordering The ordering of the channel (ORDERED or UNORDERED).
-     * @param feeEnabled Whether fees are enabled for the channel.
+     * @param feeEnabled Whether fees are enabled for the channel. Note: This value isn't currently used
      * @param connectionHops The list of connection hops associated with the channel; with the first channel in this
      * array always starting from the chain this contract is deployed on
      * @param counterparty The counterparty information of the sender
@@ -267,7 +268,8 @@ contract Dispatcher is OwnableUpgradeable, UUPSUpgradeable, ReentrancyGuard, IDi
      * @param connectionHops The list of connection hops associated with the channel, with the first channel in this
      * array always starting from the chain this contract is deployed on.
      * @param ordering The ordering of the channel (ORDERED or UNORDERED).
-     * @param feeEnabled A boolean indicating whether fees are enabled for the channel.
+     * @param feeEnabled A boolean indicating whether fees are enabled for the channel. Note: This value isn't currently
+     * used
      * @param counterparty The counterparty information for the channel.
      * @param proof The proof that the counterparty is in the ACK_PENDING state (i.e. that it responded with a
      * successful channelOpenTry )
@@ -702,7 +704,8 @@ contract Dispatcher is OwnableUpgradeable, UUPSUpgradeable, ReentrancyGuard, IDi
      * @param connectionHops The connection hops associated with the channel, with the first channel in this
      * array always starting from the chain this contract is deployed on.
      * @param ordering The ordering of the channel.
-     * @param feeEnabled A boolean indicating whether fees are enabled for the channel.
+     * @param feeEnabled A boolean indicating whether fees are enabled for the channel. Note: This value isn't currently
+     * used
      * @param counterparty The details of the counterparty.
      */
     function _connectChannel(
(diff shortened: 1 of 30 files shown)


contracts/core/Dispatcher.sol
```diff
@@ -724,12 +713,10 @@ contract Dispatcher is OwnableUpgradeable, UUPSUpgradeable, ReentrancyGuard, IDi
         bool feeEnabled,
         ChannelEnd calldata counterparty
     ) internal {
-        // Register port and channel mapping
-        // TODO: check duplicated channel registration?
-        // TODO: The call to `Channel` constructor MUST be move to `openIbcChannel` phase
-        //       Then `connectIbcChannel` phase can use the `version` as part of `require` condition.
+        // We don't need to check that a channel isn't already present between a portAddress and a chanenlId since
+        // polymer chain verification should prevent double registration of the same channel.
```
nicopernas, line 717: maybe add a note about the `feeEnabled` case here?

contracts/core/UniversalChannelHandler.sol
```diff
@@ -40,8 +32,6 @@ import {UUPSUpgradeable} from "@openzeppelin/contracts/proxy/utils/UUPSUpgradeab
  * @dev This contract can integrate directly with dapps, or a middleware stack for packet routing.
  */
 contract UniversalChannelHandler is IbcReceiverBaseUpgradeable, UUPSUpgradeable, IbcUniversalChannelMW {
-    uint256[49] private __gap;
```
nicopernas, line 43: is this safe to upgrade to?

contracts/interfaces/IbcReceiverUpgradeable.sol
```diff
@@ -57,4 +56,12 @@ contract IbcReceiverBaseUpgradeable is OwnableUpgradeable {
...
         dispatcher = _dispatcher;
     }
+
+    /**
+     * @dev This empty reserved space is put in place to allow future versions to add new
+     * variables without shifting down storage in the inheritance chain.
+     * See https://docs.openzeppelin.com/contracts/4.x/upgradeable#storage_gaps
+     */
+    // solhint-disable-next-line ordering
+    uint256[49] private __gap;
```
nicopernas, line 66: ok so the gap was moved from the other place to here?

nicopernas (approved):
Only minor comments/questions.

contracts/core/UniversalChannelHandler.sol
RnkSngh: yep. we're removing gap here but adding it to the IbcReceiverbaseUpgradeable so it should be ok!

contracts/interfaces/IbcReceiverUpgradeable.sol
RnkSngh: exactly

contracts/core/Dispatcher.sol
RnkSngh: good one - added comments, and committed it as acknowledging (but not fixing) the lack of feeEnabled validation

contracts/interfaces/IbcReceiverUpgradeable.sol
RnkSngh: the reason why it's better in the IbcreceiverUpgradeable is that any contracts which inherit from it (e.g. user Dapps) are also now going to be upgrade-safe in case we decide we want to add any storage vars to this contract

contracts/interfaces/IbcMiddleware.sol
```diff
@@ -247,3 +250,29 @@ contract IbcMwUser is Ownable {
         authorizedMws[address(middleware)] = true;
     }
 }
+
+abstract contract IbcUniversalPacketReceiverBase is IbcMwUser, IbcUniversalPacketReceiver {
```
alfredo-stonk, line 254: For clarity, should we move this base contract along with `IbcUniversalPacketReceiver` to a new file `IbcUniversalChannelReceiver.sol` ?

contracts/interfaces/IbcMiddleware.sol
```diff
@@ -247,3 +250,29 @@ contract IbcMwUser is Ownable {
...
+    modifier onlyAuthorizedChannel(bytes32 channelId) {
+        if (!authorizedChannelIds[channelId]) {
+            revert unauthorizedChannel();
+        }
+        _;
+    }
+
+    constructor(address _middleware) IbcMwUser(_middleware) {}
+
+    function _authorizeChannel(bytes32 channelId) internal {
```
alfredo-stonk, line 272: add detailed docs for what this channelID should be?
eg. it must be a UC channelID from the running chain's perspective, or better with an example, a UC with two ends, channel-1 on chain A and channel-2 on B, then contract on A should authorize channel-1, etc

alfredo-stonk (approved):

contracts/interfaces/IbcMiddleware.sol
RnkSngh: added some natspec with an example!

contracts/interfaces/IbcMiddleware.sol
RnkSngh: I agree we should do this; In general the example contracts/interfaces need some cleanup/clarifications. There's an open issue on the kanban to tack this. 

I think we should do any organization/clarity changes in a separate PR to separate out audit-specific fixes/docs from general clarity ones so I think we can do this change in another PR?

contracts/interfaces/IbcMiddleware.sol
RnkSngh: I think this specific doc is relevant to issue 23 so I added it to [that commit ](https://github.com/open-ibc/vibc-core-smart-contracts/pull/108/commits/1da74e81d86989b537a2a032406c8e31050a70c3)

dshiell (approved):
very nice!

PR #112, opened 2024-06-04
init moon contract for loadtesting

jlafiandra6:
currently this contract is a copy of the mars contract, but this will be used for load testing and we might add stat tracking or other functionality.

## Summary by CodeRabbit

- **New Features**
  - Introduced the `Moon` contract as an IBC receiver, with capabilities for handling packet reception, acknowledgments, timeouts, and channel operations.
  - Added specialized contracts for testing error handling scenarios related to IBC operations.

- **Enhancements**
  - Extended the `Moon` contract with new public state variables and functions to improve channel management and packet handling.
  - Added a constructor with an `IbcDispatcher` parameter for better initialization.

- **Test Enhancements**
  - Implemented various specialized contracts inheriting from `Moon` to test different error scenarios, including custom revert behaviors and panic conditions.


contracts/examples/Moon.sol
```diff
@@ -0,0 +1,255 @@
...
+    }
+
+    function onChanCloseInit(bytes32 channelId, string calldata, bytes32) external virtual onlyIbcDispatcher {}
+}
+
+/*
+ * These contracts are the exact same as Moon, but they revert/panick in different ways.
+ * they are used to test that transport error is seperated from errors thrown in onRecvPacket
+ */
+contract RevertingStringMoon is Moon {
```
RnkSngh, line 167: I think we can get rid of all the revertingString contracts since those are only used in unit test suite. You won't ever need to deploy these to testnets even for load testing

contracts/examples/Moon.sol
```diff
@@ -0,0 +1,255 @@
...
+import {IbcDispatcher} from "../interfaces/IbcDispatcher.sol";
+
+/**
+ * @title Moon
+ * @notice Moon is a simple IBC receiver contract that receives packets and sends acks.
+ * For now, it is a copy of Mars.sol, but may be extended in the future.
+ * @dev This contract is used for only testing IBC functionality and as an example for dapp developers
+ *  on how to integrate with the vibc protocol.
+ */
+contract Moon is IbcReceiverBase, IbcReceiver {
```
RnkSngh, line 31: Since moon is just a copy of mars right now, I'm wondering if we should use inheritance instead of copying over the whole contract so it's easier to see what the specific differences are between Moon & Mars, and to avoid having to duplicate updates that need to be done on both contracst.

For any new moon specific functionality you need to add, we can just override each method 

```suggestion
contract Moon is Mars{}

```

So all of these LOC can be changed to:

RnkSngh (changes requested):

RnkSngh (approved):

PR #114, opened 2024-06-13
Raunak/dispatcher agnostic fee implementation

RnkSngh:
PR to add a simple fee vault integration:

- add a feevault contract which holds pre-paid fees for the protocol
- added a getter and set feeVault contract in dispatcher
- modified mars, earth, and uch packet to have functionality to send packets and init channels with fees
- updated test suite to use fee functionality + added unit tests for fee vaults

## Summary by CodeRabbit

- **New Features**
  - Introduced fee-handling functionality across multiple contracts.
  - Added `sendUniversalPacketWithFee` method for fee-based packet sending.
  
- **Bug Fixes**
  - Updated packet sending functions to return sequence values for improved tracking.

- **Tests**
  - Enhanced test cases to include fee-related scenarios.
  - Implemented fuzz testing for fee calculations and withdrawals.

- **Refactor**
  - Updated multiple contracts to incorporate fee vault initialization and management.

- **Chores**
  - Improved import statements and function signatures to support new fee-handling features.

### DIFF
--- a/contracts/core/Dispatcher.sol
+++ b/contracts/core/Dispatcher.sol
@@ -28,6 +28,7 @@ import {ReentrancyGuard} from "@openzeppelin/contracts/security/ReentrancyGuard.
 import {Channel, ChannelEnd, ChannelOrder, IbcPacket, ChannelState, AckPacket, Ibc} from "../libs/Ibc.sol";
 import {IBCErrors} from "../libs/IbcErrors.sol";
 import {IbcUtils} from "../libs/IbcUtils.sol";
+import {IFeeVault} from "../interfaces/IFeeVault.sol";
 
 /**
  * @title Dispatcher
@@ -64,6 +65,7 @@ contract Dispatcher is OwnableUpgradeable, UUPSUpgradeable, ReentrancyGuard, IDi
     ILightClient _UNUSED; // From previous dispatcher version
     mapping(bytes32 => string) private _channelIdToConnection;
     mapping(string => ILightClient) private _connectionToLightClient;
+    IFeeVault public feeVault;
 
     constructor() {
         _disableInitializers();
@@ -75,13 +77,17 @@ contract Dispatcher is OwnableUpgradeable, UUPSUpgradeable, ReentrancyGuard, IDi
      * @dev This method should be called only once during contract deployment.
      * @dev For contract upgarades, which need to reinitialize the contract, use the reinitializer modifier.
      */
-    function initialize(string memory initPortPrefix) public virtual initializer nonReentrant {
+    function initialize(string memory initPortPrefix, IFeeVault _feeVault) public virtual initializer nonReentrant {
         if (bytes(initPortPrefix).length == 0) {
             revert IBCErrors.invalidPortPrefix();
+        }
+        if (address(_feeVault) == address(0)) {
+            revert IBCErrors.invalidAddress();
         }
         __Ownable_init();
         portPrefix = initPortPrefix;
         portPrefixLen = uint32(bytes(initPortPrefix).length);
+        feeVault = _feeVault;
     }
 
     /**
--- a/contracts/core/Dispatcher.sol
+++ b/contracts/core/Dispatcher.sol
@@ -473,8 +473,14 @@ contract Dispatcher is OwnableUpgradeable, UUPSUpgradeable, ReentrancyGuard, IDi
      * @param packet The packet data to send.
      * @param timeoutTimestamp The timestamp, in seconds after the unix epoch, after which the packet times out if it
      * has not been received.
+     * @return sequence The sequence number of the packet, starting from 1 and incremented with each sendPacket. This
+     * sequence number is used to link with depositing packet fees in the fee vault
      */
-    function sendPacket(bytes32 channelId, bytes calldata packet, uint64 timeoutTimestamp) external nonReentrant {
+    function sendPacket(bytes32 channelId, bytes calldata packet, uint64 timeoutTimestamp)
+        external
+        nonReentrant
+        returns (uint64 sequence)
+    {
         // ensure port owns channel
         if (_portChannelMap[msg.sender][channelId].counterpartyChannelId == bytes32(0)) {
             revert IBCErrors.channelNotOwnedBySender();
@@ -484,7 +490,7 @@ contract Dispatcher is OwnableUpgradeable, UUPSUpgradeable, ReentrancyGuard, IDi
         }
 
         // current packet sequence
-        uint64 sequence = _nextSequenceSend[msg.sender][channelId];
+        sequence = _nextSequenceSend[msg.sender][channelId];
         if (sequence == 0) {
             revert IBCErrors.invalidPacketSequence();
         }
--- a/contracts/core/FeeVault.sol
+++ b/contracts/core/FeeVault.sol
@@ -1,0 +1,91 @@
+// SPDX-License-Identifier: Apache-2.0
+/*
+ * Copyright 2024, Polymer Labs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+pragma solidity 0.8.15;
+
+import {Ownable} from "@openzeppelin/contracts/access/Ownable.sol";
+import {IFeeVault} from "../interfaces/IFeeVault.sol";
+import {ReentrancyGuard} from "@openzeppelin/contracts/security/ReentrancyGuard.sol";
+import {ChannelOrder} from "../libs/Ibc.sol";
+
+contract FeeVault is Ownable, ReentrancyGuard, IFeeVault {
+    /**
+     * @notice Deposits the send packet fee for a given channel and sequence that is used for relaying recieve and
+     * acknowledge steps of a packet handhsake after a dapp has called the sendPacket on dispatcher.
+     * @dev This function calculates the required fee based on provided gas limits and gas prices,
+     *      and reverts if the sent value does not match the calculated fee.
+     *      The first entry in `gasLimits` and `gasPrices` arrays corresponds to `recvPacket` fees,
+     *      and the second entry corresponds to `ackPacket` fees.
+     * @param channelId The identifier of the channel.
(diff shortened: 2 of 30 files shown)


RnkSngh:
Note: we're still missing the last implementation detail of how we can identify channel open fees

test/FeeVault.t.sol
```diff
@@ -0,0 +1,83 @@
...
+    function testFuzz_sendPacket_FeesAddUpToValue(
+        uint104 gasFee1,
+        uint104 gasFee2,
+        uint104 gasLimit1,
+        uint104 gasLimit2,
+        bool shouldRevert,
+        uint56 fuzz
+    ) public {
+        gasFee1 = uint104(bound(gasFee1, 1, 2 ** 104 - 1));
+        gasLimit1 = uint104(bound(gasFee1, 1, 2 ** 104 - 1));
```
dshiell, line 31: `gasLimit1 = gasFee1` ?

test/upgradeableProxy/DispatcherRC4.upgrade.t.sol
```diff
@@ -35,15 +25,43 @@
...
+    Earth as EarthRc4,
+    UniversalPacket as UniversalPacketRc4,
+    AckPacket as AckPacketRc4,
+    IbcUtils
+} from "./upgrades/EarthRc4.sol";
+
+abstract contract DispatcherRC4TestUtils is ChannelHandShakeUpgradeUtil {
+    function sendOnePacketRc4(bytes32 channelId, uint64 packetSeq, MarsRc4 sender) public {
+        vm.expectEmit(true, true, true, true);
+        emit SendPacket(address(sender), channelId, packet, packetSeq, timeoutTimestamp);
```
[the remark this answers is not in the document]
dshiell: what is `packet` actually set to here?

dshiell (commented):
would be good to break large PRs like this into smaller ones in future. for example, changes for adding sequence number return could be separate from fee vault changes. in this case fee fault addition alone would be rather large, probably stackable prs would be perfect here.

test/upgradeableProxy/DispatcherRC4.upgrade.t.sol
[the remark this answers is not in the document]
RnkSngh: packet is set from the inherited contract; [here](https://github.com/open-ibc/vibc-core-smart-contracts/blob/6f40816d6fa9f4572868a263b42a075323552b38/test/upgradeableProxy/Dispatcher.upgrade.t.sol#L86) it is set to the encoded string "msgPayload" 

I updated this lineuto be the hardcoded string in case that's better testing?

contracts/examples/Mars.sol
```diff
@@ -37,6 +39,7 @@
     bytes32[] public connectedChannels;
 
     string[] public supportedVersions = ["1.0", "2.0"];
+    IFeeVault public feeVault;
```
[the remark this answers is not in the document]
RnkSngh: ah, forgot to delete this line, should be deleted now

contracts/core/FeeVault.sol
```diff
@@ -0,0 +1,52 @@
...
+
+    function depositOpenChannelFee(address src, string calldata destPortId, string[] calldata connectionHops)
+        external
+        payable
+        nonReentrant
+    {
+        if (msg.value == 0) {
+            revert NoFeeSent();
+        }
+        emit OpenChannelFeeDeposited(src, destPortId, connectionHops);
```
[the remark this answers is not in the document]
RnkSngh: > I think we need to emit all the parameters sent to chanopeninit

so you're saying, instead of just the src address and dest portId, we also need version & ordering to for our 
unique key value? 

> shouldn't we also emit the amount of fees paid?

ah yes, added now :/

contracts/core/UniversalChannelHandler.sol
```diff
@@ -82,12 +83,38 @@ contract UniversalChannelHandler is IbcReceiverBaseUpgradeable, UUPSUpgradeable,
...
+    function sendUniversalPacketWithFee(
+        bytes32 channelId,
+        bytes32 destPortAddr,
+        bytes calldata appData,
+        uint64 timeoutTimestamp,
+        uint256[2] calldata gasLimits,
+        uint256[2] calldata gasPrices
+    ) external payable returns (uint64 sequence) {
+        // Cache dispatcher for gas savings
+        IbcDispatcher _dispatcher = dispatcher;
```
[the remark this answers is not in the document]
RnkSngh: > should we do this with sendUniversalPacket then?

by "this" you mean caching for gas savings yes? 

if so,  we don't do this with `sendUniversalPacket` because there is only one storage read, whereas with `sendUniversalPacketWithFee` there (would have been) two potential storage reads to find the dispatcher address, one to send the packet and the second to query the feeVault. 

Storage reads are more expensive than a memory read + write, so it becomes worth it to cache the dispatcher address in memory when we have more than one storage read. But if we only have one storage read (as in the case of `sendUniversalPacket`, caching is more expensive than just doing the storage read (i.e. it's one storage read vs 1 storage read + 1 memory write)

contracts/core/UniversalChannelHandler.sol
```diff
@@ -185,6 +212,17 @@ contract UniversalChannelHandler is IbcReceiverBaseUpgradeable, UUPSUpgradeable,
 
     function _authorizeUpgrade(address newImplementation) internal override onlyOwner {}
 
+    function _sendUCHPacket(bytes32 channelId, bytes32 destPortAddr, bytes calldata appData, uint64 timeoutTimestamp)
```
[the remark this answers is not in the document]
RnkSngh: ah yes, this was an internal function that was left over from the many iterations - should be removed now!

test/FeeVault.t.sol
RnkSngh: now that I'm revisiting this, we can actually get rid of these bounds altogether

RnkSngh: yep, thanks!

contracts/core/UniversalChannelHandler.sol
```diff
@@ -82,12 +83,38 @@ contract UniversalChannelHandler is IbcReceiverBaseUpgradeable, UUPSUpgradeable,
...
+        uint256[2] calldata gasPrices
+    ) external payable returns (uint64 sequence) {
+        // Cache dispatcher for gas savings
+        IbcDispatcher _dispatcher = dispatcher;
+
+        bytes memory packetData = IbcUtils.toUniversalPacketBytes(
+            UniversalPacket(IbcUtils.toBytes32(msg.sender), MW_ID, destPortAddr, appData)
+        );
+        emit UCHPacketSent(msg.sender, destPortAddr);
+        sequence = _dispatcher.sendPacket(channelId, packetData, timeoutTimestamp);
```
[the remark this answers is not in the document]
RnkSngh: it would be slightly less gas efficient if they both used the same method - one caches the dispatcher for gas, the other doesn't

dshiell (approved):

dshiell (changes requested):
checks tests?

contracts/core/Dispatcher.sol
```diff
@@ -474,7 +477,11 @@ contract Dispatcher is OwnableUpgradeable, UUPSUpgradeable, ReentrancyGuard, IDi
      * @param timeoutTimestamp The timestamp, in seconds after the unix epoch, after which the packet times out if it
      * has not been received.
      */
-    function sendPacket(bytes32 channelId, bytes calldata packet, uint64 timeoutTimestamp) external nonReentrant {
+    function sendPacket(bytes32 channelId, bytes calldata packet, uint64 timeoutTimestamp)
```
nicopernas, line 480: nit: update the docs with the new returned value?

contracts/core/FeeVault.sol
```diff
@@ -0,0 +1,52 @@
...
+import {ReentrancyGuard} from "@openzeppelin/contracts/security/ReentrancyGuard.sol";
+
+contract FeeVault is Ownable, ReentrancyGuard, IFeeVault {
+    function depositSendPacketFee(
+        bytes32 channelId,
+        uint64 sequence,
+        uint256[2] calldata gasLimits,
+        uint256[2] calldata gasPrices
+    ) external payable nonReentrant {
+        uint256 fee = gasLimits[0] * gasPrices[0] + gasLimits[1] * gasPrices[1];
```
nicopernas, line 31: I guess we are using `uint256[2]` to save gas but can we document what these values mean? maybe a comment next to `gasLimits[0] * gasPrices[0] + gasLimits[1] * gasPrices[1];` explaining the reasoning behind would help the reader.

contracts/core/FeeVault.sol
```diff
@@ -0,0 +1,52 @@
...
+        }
+        emit SendPacketFeeDeposited(channelId, sequence, gasLimits, gasPrices);
+    }
+
+    function depositOpenChannelFee(address src, string calldata destPortId, string[] calldata connectionHops)
+        external
+        payable
+        nonReentrant
+    {
+        if (msg.value == 0) {
```
nicopernas, line 43: Don't we have a flat fee for opening channels?

contracts/examples/Mars.sol
```diff
@@ -91,14 +104,36 @@ contract Mars is IbcReceiverBase, IbcReceiver {
...
     }
 
+    /**
+     * @dev Sends a packet with a greeting message over a specified channel.
+     * @param message The greeting message to be sent.
+     * @param channelId The ID of the channel to send the packet to.
+     * @param timeoutTimestamp The timestamp at which the packet will expire if not received.
+     * @dev This method also returns sequence from the dispatcher for easy testing
+     */
+    function greet(string calldata message, bytes32 channelId, uint64 timeoutTimestamp)
```
nicopernas, line 114: Does this need to be `payable` too?

contracts/core/FeeVault.sol
nicopernas: I just saw your comment on this.

contracts/core/Dispatcher.sol
RnkSngh: ah yes, added - good catch!

contracts/core/FeeVault.sol
RnkSngh: added this and natspec in general for the fee vault, again good catch!

contracts/examples/Mars.sol
RnkSngh: no, because greet should only be used when there isn't a fee. greetWithFee is what should be used if there needs to be a fee sent, and that is indeed payable. 

I added docs clarifying this since i'd imagine this might be a source of confusion in general

nicopernas: yeah, that's what I thought too. note that it is marked as `payable` though - hence the question :D

RnkSngh: ahhh whoops didn't realize it was payable

RnkSngh: should be made nonpayable now!

dshiell (approved):
lgtm!

PR #131, opened 2024-06-27
h5law/release-workflows

h5law:
## Summary by CodeRabbit

- **Chores**
  - Updated workflows to use the latest versions of tools and actions, including Bun v2.
  - Improved workflow steps for generating and checking bindings.

- **New Features**
  - Introduced Go bindings for various Ethereum contracts, enabling easier interaction with these contracts via auto-generated methods and structs. This includes contracts like `ProtoChannel`, `DummyLightClient`, `FeeVault`, and many more.

- **Refactor**
  - Enhanced Makefile to dynamically handle contract bindings generation and added new targets for generating Go and TypeScript bindings.

- **Documentation**
  - Added and updated metadata for various Ethereum contract bindings to enhance clarity and usability.


go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/open-ibc/vib-core-smart-contracts/bindings
+module github.com/open-ibc/vib-core-smart-contracts/v2
```
RnkSngh, line 1: v2 might be confusing since that is a different number than our tagged versions?

bun.lockb
RnkSngh, line 1: let's  add the bun-related files to our .npmignore so that they aren't published in the npm package

go.mod
h5law: This has to be the major version only and you specify with a postfix of `"@v2.1.0"` or another full semantic version when doing go get/install/list etc the full version string [1]. The need to have a separate `/v2` directory is not needed as displayed by numerous packages and the go docs [2].

[1] https://go.dev/blog/v2-go-modules
[2] https://go.dev/wiki/Modules#releasing-modules-v2-or-higher

RnkSngh: what happens when we upgrade to major version v3? would we have to manually change this url then/remember to change it?

h5law: Yeah the suffix would change to `/v3` CI would fail with a `v2` suffix and `v3.x.x` tag so will force us to remember when we do a tag release.

go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/open-ibc/vib-core-smart-contracts/bindings
+module github.com/polymerdao/open-ibc/vibc-core-smart-contracts/v2
```
nicopernas, line 1: Shouldn't this be `github.com/open-ibc/vibc-core-smart-contracts/xxx`?
Also, `v2` should be `go-bindings` or similar. It's ultimately the package name and should not include versioning, you have the package version for that :)

nicopernas:
I'm clearly late since this is in `main` already but the go module should not live in the `bindings` directory. That's too generic. It should be something like `go-bindings` since I'd expect another directory called `ts-bindings` and even `py-bindings`

go.mod
h5law: Yeah so the v2 is needed once the major version >= 2.x.x according to the go documentation for modules [1]. With regards to moving the bindings to language specific bindings I don't mind - currently ts-bindings will come from the npm package I think and live in the `src` directory so it'd be nice to move this out into its own directory but I am unsure of its need at this moment. I'd like them all to live in the bindings directory something like `bindings/{go,ts,py}` for example.

[1]] https://github.com/open-ibc/vibc-core-smart-contracts/pull/131#discussion_r1659149330

h5law:
@nicopernas I have moved all bindings into the `./bindings` directory which can be generated by `make bindings-gen-go bindings-gen-ts clean` (requires `abigen` and `typechain`) the typescript bindings are copied on generation into `./src/evm/contracts` for ease of use in the npm package (the `./bindings/{go,ts}` directories are excluded from the npm package itself (except the copy of `ts` made previously). I also brought the go bindings in line with the typescript ones in terms of scope and what is being made accessible throught the API.

Let me know if there is anything else.

h5law:
@coderabbitai ignore

h5law:
@coderabbitai resolve

.github/workflows/bindings.yml
```diff
@@ -25,30 +25,22 @@ jobs:
...
 
-      - name: Check for Go Bindings Changes
+      - name: Check for Bindings Changes
         run: |
           if [ $(git status --porcelain | sed 's/^...//' | grep bindings/ | wc -l) -ne 0 ]; then
-            echo "Changes in Go Bindings"
+            echo "Changes in Generated Bindings"
             echo $(git status --porcelain | sed 's/^...//' | grep bindings/)
             exit 1
           fi
```
nicopernas, line 46: nit: you could rewrite this as follows 
```bash
changes=$( git status --porcelain -- ./bindings )
if [ -n "$changes" ]; then
  echo -e "Changes in Generated Bindings:\n$changes"
  exit 1
fi
```

Makefile
```diff
@@ -1,15 +1,32 @@
 .SILENT:
 
-CONTRACT_ABI_FILES = \
-	./out/Dispatcher.sol/Dispatcher.json \
-	./out/UniversalChannelHandler.sol/UniversalChannelHandler.json \
-	./out/Mars.sol/Mars.json \
-	./out/Earth.sol/Earth.json \
-	./out/Moon.sol/Moon.json
+# Define the list of contract names
+CONTRACT_NAMES = OptimisticProofVerifier \
```
nicopernas, line 4: could we use a list or array to avoid the `\` ?

Makefile
```diff
@@ -21,20 +38,46 @@ extract_artifacts: build-contracts
...
 	mkdir -p $(TMP_ABI_DIR); \
 	for abi_file in $(CONTRACT_ABI_FILES); do \
-		cat $$abi_file | jq .abi > $(TMP_ABI_DIR)/$$(basename $$abi_file); \
+		dir=$$(basename $$(dirname $$abi_file)); \
+		mkdir -p $(TMP_ABI_DIR)/$$dir; \
+		cat $$abi_file | jq .abi > $(TMP_ABI_DIR)/$$dir/$$(basename $$abi_file); \
 	done
 
-abigen: extract_artifacts
+# Libraries do not generate the correct ABI code (ChannelState enum causes errors)
```
nicopernas, line 46: nice!

nicopernas (approved):
Left a couple of nits. Looks good otherwise!

h5law:
@coderabitai resume

h5law:
https://github.com/coderabbitai resume

h5law:
@coderabbitai resume

h5law:
@coderabbitai ignore

RnkSngh (approved):
approving for now so that we can start integrating with polymerase - feel free to save these comments for another PR

.github/workflows/bindings.yml
```diff
@@ -25,30 +25,22 @@ jobs:
...
-            exit 1
-          fi
-
-      - name: Check for TypeScript Bindings Changes
+      - name: Check for Bindings Changes
         run: |
-          if [ $(git status --porcelain | sed 's/^...//' | grep src/evm/contracts/ | wc -l) -ne 0 ]; then
-            echo "Changes in TypeScript Bindings"
-            echo $(git status --porcelain | sed 's/^...//' | grep src/evm/contracts/)
+          changes=$( git status --porcelain -- ./bindings )
```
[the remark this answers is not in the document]
h5law: My mistake this was before I moved them out I will add that in now

.github/workflows/release.yml
```diff
@@ -80,20 +95,31 @@ jobs:
...
+      - name: Update Go Registry
+        run: |
+          GOPROXY=proxy.golang.org go list \
+          -m github.com/open-ibc/vibc-core-smart-contracts/v2@${{ steps.version.outputs.release }}
+
       - name: Update NPM Registry
         run: |
-          npm ci
-          npm publish --provenance --access public
+          bun install --frozen-lockfile
```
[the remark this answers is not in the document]
h5law: This is the equivalent of `npm ci` but it was able to work easier than messing around with npm.

PR #133, opened 2024-07-02
fix contract spec to work with fee vault dispatcher

RnkSngh:
This is a quick PR to fix the default spec exported in the npm package, so that it correctly initializes dispatcher proxy with the new implementation (which takes in an additional feeVaultAddress

## Summary by CodeRabbit

- **New Features**
	- Updated the `DispatcherProxy` contract to include an additional argument for initialization, allowing users to specify an address for `FeeVault`. This enhances the contract's flexibility during setup.


h5law (approved):
LGTM lets :shipit:

PR #134, opened 2024-07-02
make dispatcher reinitializable + add initilization to upgrade specs

RnkSngh:
PR to make dispatcher reinitializable 

this adds a minor change to our dispatcher implementation, where we have our initializing function inherit from `reinitializer(2)` rather than initializer. This will allow us to override the init version for our  existing prod + staging contracts to allow setting the fee vault. 

Correspondingly, there is a minor feature addition to add initialization to trasnactions as well (similar to existing functionality of constructor initialization), which renders $INITDATA sent to tx specs. This will be utilized in proxy upgrades which utilize initializers (i.e. any contract upgrades which need to modify setting state) 

## Summary by CodeRabbit

- **New Features**
	- Improved initialization logic for deploying smart contracts.
	- Enhanced transaction schema with new initialization fields.
	- Added support for encoding initialization arguments using the `ethers` library.
- **Bug Fixes**
	- Fixed visibility and reinitialization issues in the Dispatcher contract.
- **Updates**
	- Modified the initialization parameters in deployment and transaction functions for better consistency and clarity.


Inkvi (approved):

h5law (approved):
LGTM small but good change - curious does this not change the bindings?

RnkSngh:
> LGTM small but good change - curious does this not change the bindings?

We didn't change the interface (making a function external vs public doesn't change the abi at all) so no go bindings were changed. 

We DID change the contract bytecode, which is why we have new typescript bindings

PR #154, opened 2024-07-29
Misc Contract Finalizations

RnkSngh:
PR to add the following minor tweaks (note, proxy compatibility tests would break after adding 2 step ownable, so they're both modified in this same PR): 
- add 2 step ownable to dispatcher, uch, and fee vault
- Move to dispathcer inheriting from upgradeable reentrnacy guard rather than just reentrnacy guard (also was something we were holding off on doing for storage compatibility) 
- Remove upgradeable proxy tests - since we're no longer prioritizing unchanging proxy addresses

## Summary by CodeRabbit

- **New Features**
	- Enhanced security with a two-step ownership transfer mechanism across multiple contracts.
	- Improved packet handling and acknowledgment processes for better traceability.
- **Bug Fixes**
	- Adjusted constants in connection and packet commitment handling to align with the updated protocol.
- **Refactor**
	- Removed legacy functionality to streamline the upgrade processes in the Dispatcher and related contracts.
- **Tests**
	- Updated tests to reflect changes in packet sending and internal logic, ensuring current functionality is validated.


dshiell (approved):
lgtm!

PR #156, opened 2024-07-30
Add missing release artifacts to help with verification

dshiell:
## Summary by CodeRabbit

- **Improvements**
	- Enhanced artifact collection process by including additional JSON files in the output directory.
	- Standardized string formatting for better consistency in workflow configurations.

PR #157, opened 2024-07-31
add missing release artifacts to help with verification & only build …

RnkSngh:
PR to add _all_ contracts in our contracts folder (i.e. skipping our foundry test contracts) in our artifiacts. 

Also, this workflow renames the artifacts folder to release-artifacts, since artifacts is also a folder which can be written to by foundry during compile

this is based on @dshiell 's pr [here](https://github.com/open-ibc/vibc-core-smart-contracts/pull/156)

## Summary by CodeRabbit

- **Chores**
	- Improved the GitHub Actions workflow for releases, enhancing consistency and clarity in file management.
	- Implemented a cleanup step before building contracts to prevent issues with stale artifacts.
	- Updated artifact paths for better organization of output files.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,7 +3,7 @@ name: Release
 on:
   push:
     tags:
-      - "v*.*.*"
+      - 'v*.*.*'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.ref_name }}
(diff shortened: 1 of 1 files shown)


dshiell (approved):
nice! let's see how crazy the releases look haha, possibly we may want to tarball the contracts.

PR #158, opened 2024-07-31
remove onChanOpenInit and onChanCloseInit callbacks from dispatcher, UCH, and example dapps

RnkSngh:
PR to remove the onChanOpenInit and onChanCloseInit callbacks from dispatcher, since with our new open channel access control, this method is redundant. 

OnChanOpeninit in ibc spec is used for version negotiation. But since only the dapp should be able to open it's own channel, the dapp should just be able to pass in the version it wants to use. 

This should result in some minor contract size decrease and decreased gas costs for the first step of the channel handshake 

We are also removing the onChanCloseInit since the dapp can always query the coutnerparty information from the dispatcher. This is more expensive (15k gas) for the dapp to query, but If we find that dapps do need this information, we can always modify the chanCloseInit method to return this information, and that would be a non breaking change

## Summary by CodeRabbit

- **New Features**
  - Enhanced the `Dispatcher` contract functionality with improved channel event handling.

- **Bug Fixes**
  - Removed obsolete functions related to channel management, reducing potential errors.

- **Documentation**
  - Updated comments in the `Dispatcher` contract to clarify functionality and improve understanding.

- **Chores**
  - Cleaned up code by removing unnecessary commented sections and ensuring clarity in interfaces.


dshiell (approved):
lgtm!

nicopernas (approved):

alfredo-stonk:
Context on why these IBC callback methods exist in the first place:
In our initial design, any Relayer EOA can call Dispatcher's `openChannelInit` or `closeChannelInit` methods on dApp's behalf, in which case callback methods provide dApp's access control logic. Without the callbacks, anyone can create new or close existing channels for any dApp. This aligns with [ibc specs](https://github.com/cosmos/ibc/blob/main/spec/core/ics-004-channel-and-packet-semantics/). 

Recently, to simplify dApp logic, we require that only (or would-be) channel owner dApps can initial calls to Dispatcher for creating new and closing existing channels. `onOpenChanInit` and `onCloseChanInit` methods are no longer necessary. 

If dApps need to perform extra logic after calls to Dispatcher, they can do so without callback, eg.
```solidity
function triggerChanClose(bytes32 channelID) {
  dispatcher.closeChannelInit(channelID)
  // extra cleanup logic 
 cleanup(channelID)
}
```

contracts/core/Dispatcher.sol
```diff
@@ -408,24 +392,11 @@ contract Dispatcher is Ownable2StepUpgradeable, UUPSUpgradeable, ReentrancyGuard
...
         }
-        (bool success, bytes memory data) = _callIfContract(
-            msg.sender,
-            abi.encodeWithSelector(
-                IbcChannelReceiver.onChanCloseInit.selector,
-                channelId,
-                channel.counterpartyPortId,
-                channel.counterpartyChannelId
-            )
-        );
```
alfredo-stonk, line 395: nit: can we refactor channel ownership check into a common method? The check is done in multiple places.  Now the full channel is no longer required after `onChanCloseInit` is removed.

alfredo-stonk (approved):

contracts/core/Dispatcher.sol
RnkSngh: ah good one. This check is done in 3 different places - sendpacket, channelCloseInit, and channelCloseConfirm. 

Channel close confirm requires the channel afterwards, so to avoid adding an extra SLOAD in channelCloseConfirm or forcing sendpacket & channelCloseInit to call a method which returns more data than it needs to (increasing gas cost considerably), I didn't have this use the common method.

We can use the same method for sendpacket and channelCloseInit though! FYI, this does seem to add an extra 40 bytes in dispatcher contract size (it seems like the code for the method itself is larger than encoding/decoding the data into an internal function), but this is _probably_ negligible since we're now working with 1600 bytes of room in our dispatcher.

alfredo-stonk (approved):

PR #166, opened 2024-08-12
read accounts into env during deployment, to allow referincing to acc…

RnkSngh:
this PR reads all accounts from the accounts.spec file into the env. This allows us to reference multisig safe addresses from accounts in the account spec, rather than having to hard code everything. Specifically, this is useful for knowing which account to transfer ownership to during ownership transfers


src/utils/io.ts
```diff
@@ -239,6 +239,20 @@ export async function writeDeployedContractToFile(
   });
 }
 
+// Read existing accounts into env
+export async function readAccountsIntoEnv(
+  env: any,
+  accountRegistry: AccountRegistry
+) {
+  console.log(env, accountRegistry.keys());
```
dshiell, line 247: pub keys right?

dshiell (approved):
lgtm!

src/utils/io.ts
RnkSngh: err whoops, yes it only logs pub keys, but I should get rid of this - good catch!

RnkSngh: removed now!

PR #167, opened 2024-08-12
add nonce caching to multisig txs

RnkSngh:
Add nonce caching to multisig txs. This allows us to submit concurrent txs for our multisig without waiting for the previous one to be executed.


dshiell (approved):
lgtm!

PR #192, opened 2024-09-09
Raunak/informal q3 fixes

RnkSngh:
PR to address findings in the informal audit:
- Remove channelId to connection mapping in chanCloseInit & chanCloseConfirm to avoid cases where packets can be recieved even after a channel close has been initiated
- add version != 0 check to revert in chanCloseInit instead of in peptide


.github/workflows/bindings.yml
```diff
@@ -45,13 +45,13 @@ jobs:
           fi
       - name: Upload ts bindings artifact on error
         if: failure()
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@v3
```
nicopernas, line 48: maybe use `v4` instead? https://github.blog/changelog/2024-02-13-deprecation-notice-v1-and-v2-of-the-artifact-actions/

RnkSngh: udpated!

PR #198, opened 2024-09-17
Raunak/generic light client

RnkSngh:
PR to make our light client interfaces more generic to support non op-stack clients

Note: we'll have to update our relayer to abi.encode the proof before we send client updates

## Summary by CodeRabbit

- **New Features**
	- Introduced the `AppStateVerifier` contract for verifying proofs related to Optimistic Rollup state updates.
	- Enhanced client state management with the new `updateClient` method, allowing for more flexible proof formats.
	- Added methods for verifying membership and non-membership of key-value pairs.

- **Bug Fixes**
	- Improved handling of state proofs in tests to ensure correct serialization and processing.

- **Refactor**
	- Renamed functions from `addOpConsensusState` to `updateClient` across multiple contracts and interfaces for clarity.
	- Updated parameter types in various functions to accept more generic data formats.
	- Restructured interfaces to better separate responsibilities between application state verification and proof verification.

- **Documentation**
	- Updated comments and documentation for improved clarity and accuracy.


contracts/core/OptimisticLightClient.sol
```diff
@@ -48,18 +48,21 @@ contract OptimisticLightClient is ILightClient {
...
+        returns (uint256 fraudProofEndTime, bool ended)
+    {
         uint256 hash = consensusStates[height];
         if (hash == 0) {
             // if this is a new apphash we need to verify the provided proof. This method will revert in case
             // of invalid proof.
             verifier.verifyStateUpdate(
-                l1header, proof, bytes32(appHash), l1BlockProvider.hash(), l1BlockProvider.number()
+                l1header,
+                abi.decode(proof, (OpL2StateProof)),
```
i-norden, line 62: Where does `abi.decode` come from? `abi` is a builtin in solidity that uses the calling contract's ABI to do the decoding/encoding?

i-norden (commented):
This looks like the right approach to me! One question for my own benefit

contracts/core/OptimisticLightClient.sol
RnkSngh: yep it's a builtin!

RnkSngh: [here are the sol docs on this](https://docs.soliditylang.org/en/develop/cheatsheet.html#abi-encoding-and-decoding-functions) in case you wanted more details (admittedly they're not the easiest to find)

RnkSngh: so the `abi.decode` here is just taking the generic proof byte array and returning an `OpL2StateProof` struct

we can add a test case for this, but it should throw an error if the byte array isn't a valid `OpL2StateProof` struct

contracts/core/Dispatcher.sol
```diff
@@ -681,17 +680,15 @@ contract Dispatcher is Ownable2StepUpgradeable, UUPSUpgradeable, ReentrancyGuard
...
-     * @param height The height of the consensus state.
+     * @notice Retrieves the optimistic consensus state for a specified height and client ID.
+     * @param height The height of the apphash to query
      * @param connection The connection the client corresponds to; used to find the light client.
-     * @return appHash The application hash of the consensus state.
-     * @return fraudProofEndTime The end time of the fraud proof.
-     * @return ended A boolean indicating whether the consensus state has ended.
+     * @return appHash The apphash at the given height
      */
     function getOptimisticConsensusState(uint256 height, string calldata connection)
```
nicopernas, line 688: Should this also be `getState()` ?

contracts/core/OptimisticLightClient.sol
```diff
@@ -48,83 +51,94 @@ contract OptimisticLightClient is ILightClient {
...
      * @inheritdoc ILightClient
      */
-    function addOpConsensusState(
-        L1Header calldata l1header,
-        OpL2StateProof calldata proof,
-        uint256 height,
-        uint256 appHash
-    ) external override returns (uint256 fraudProofEndTime, bool ended) {
-        uint256 hash = consensusStates[height];
+    function updateClient(bytes calldata proof, uint256 peptideHeight, uint256 peptideAppHash) external override {
```
nicopernas, line 54: `peptide` is only our execution engine. I would suggest we use `polymerHeight` here which is more generic and does not reveal internal details of our implementation.

contracts/core/OptimisticProofVerifier.sol
```diff
@@ -132,69 +133,4 @@ contract OptimisticProofVerifier is IProofVerifier {
             revert InvalidAppHash();
         }
     }
-
```
nicopernas, line 135: nice!

contracts/interfaces/ILightClient.sol
```diff
@@ -16,7 +16,14 @@
  */
 pragma solidity ^0.8.0;
 
-import {Ics23Proof, L1Header, OpL2StateProof} from "./IProofVerifier.sol";
+import {Ics23Proof} from "./IProofVerifier.sol";
+
+enum LightClientType {
+    DummyLightClient, // Note: not deployed on any mainnets
```
nicopernas, line 22: nit: Should we use something like `SimTest` or similar instead of `Dummy`?

nicopernas (approved):
I left some comments / questions but nothing blocking. Good stuff!

contracts/core/Dispatcher.sol
RnkSngh: ah good catch, we can re-name

contracts/core/Dispatcher.sol
```diff
@@ -117,18 +117,17 @@ contract Dispatcher is Ownable2StepUpgradeable, UUPSUpgradeable, ReentrancyGuard
     /**
      * @notice Updates the client with optimistic consensus state. The optimistic consensus is accepted and will be open
      * for verification in the fraud proof window
-     * @dev Calls lightClient.addOpConsensusState; See Lightclient natspec for params information
+     * @dev Calls lightClient.updateClient; See Lightclient natspec for params information
      * @dev This function updates the client with optimistic consensus state.
      *      It should be called after verifying the optimistic consensus state on the main chain.
      */
     function updateClientWithOptimisticConsensusState(
```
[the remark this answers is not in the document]
RnkSngh: ah good catch, we can re-name

contracts/core/OptimisticLightClient.sol
RnkSngh: though if we have a different execution engine, won't we have a different light client contract? I was guessing that OptimisticLightClient is tied one-to-one with peptide hence the naming

contracts/interfaces/ILightClient.sol
RnkSngh: done!

contracts/core/OptimisticLightClient.sol
RnkSngh: ahh ok, I guess this isn't necessarily true that we'd have to add a newlight client in this case, I changed!

PR #199, opened 2024-09-17
Raunak/sequencer signer light client

RnkSngh:
## Summary by CodeRabbit

- **New Features**
	- Introduced `getLightClientType` and `consensusStates` functions for the `DummyLightClient` contract.
	- Added `SequencerSignatureVerifier` contract for validating sequencer signatures in state updates.
	- Implemented `SequencerSoloClient` for enhanced blockchain interactions with light client functionality.
	- Created `ISignatureVerifier` interface for abstracting proof verification logic.
- **Bug Fixes**
	- Enhanced proof loading mechanism in multiple tests to associate proofs with specific light clients.
- **Tests**
	- Added comprehensive tests for sequencer signature verification scenarios.
	- Updated existing tests to handle new channels and connection hops in the IBC protocol.
	- Improved test setup for loading proofs with specific light client addresses.


contracts/core/SequencerProofVerifier.sol
```diff
@@ -0,0 +1,133 @@
...
+    function _verifySequencerSignature(
+        uint256 L2BlockNumber,
+        bytes32 AppHash,
+        bytes32 L1BlockHash,
+        bytes32 L1StateRoot,
+        bytes calldata signature
+    ) internal view {
+        // TODO: check that erecover should indeed ensure that the signature signs over the hash
+        if (
+            ECDSA.recover(keccak256(bytes.concat(bytes32(L2BlockNumber), AppHash, L1BlockHash, L1StateRoot)), signature)
```
i-norden, line 93: Nice!

contracts/core/SequencerProofVerifier.sol
```diff
@@ -0,0 +1,133 @@
...
+    // @notice known L2 Output Oracle contract address to verify state update proofs against
+    address public immutable sequencer;
+
+    error InvalidSequencerSignature();
+
+    constructor(address sequencer_) {
+        sequencer = sequencer_;
+    }
+
+    function verifyStateUpdate(L1Header calldata l1header, bytes32 appHash, uint256 l2Height, bytes calldata signature)
```
i-norden, line 44: Can't we provide just the L1 blockhash here instead of the full header and hashing the full header on-chain? I don't think the header fields are used for anything else.

i-norden:
This also made me realize that I don't think we need the L1StateRoot in the SignedData. 

When doing origin consistency check in the future we will do it by either proving that the L1BlockHash in the SignedData is the same blockHash that the destination chain has in its L1BlockInfo contract or that it is an ancestor of the destination chain L1BlockInfo (provide headers linking back from the L1BlockInfo in destination chain to the L1BlockHash in peptide's SignedData) or by providing that the EIP-4788 contract on the destination EVM chain has as a member a beacon root that commits to the L1BlockHash in the SignedData).

I don't think we need to access the L1StateRoot in either case.

contracts/core/SequencerProofVerifier.sol
RnkSngh: technically yes we can and should have this method only take the blockhash. But this is only to save on how much calldata we pass in to this external `verifyStateUpdate` method. 

However, we'd still be computing the blockhash on-chain in the parent `SequencerSoloClient.updateClient`. I think that method needs to take the full header to conform to the `ILightClient` interface?

RnkSngh: we could fetch it from the l1BlockInfo contract, but that would mean that the peptide sequencer needs to be in sync with the evm chains no?

RnkSngh:
sure thing, can get rid of the `L1StateRoot`

contracts/core/SequencerProofVerifier.sol
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    /**
+     * @inheritdoc ISignatureVerifier
+     * @dev verifies a chain of ICS23 proofs
+     * Each computed subroot starting from index 0 must match the value of the next proof (hence chained proofs).
+     * The cosmos SDK and ics23 support chained proofs to switch between different proof specs.
+     * Custom proof specs are not supported here. Only Iavl and Tendermint or similar proof specs are supported.
+     */
+    function verifyMembership(bytes32 appHash, bytes calldata key, bytes calldata value, Ics23Proof calldata proofs)
```
dshiell, line 68: Seems like this and the verifyNonMembership functions could be extracted into a library to be reused across all the proof verifiers?

contracts/core/SequencerProofVerifier.sol
dshiell: Do we need to check anywhere that this update is consistent with previous state or do we trust each update in isolation because it was signed by our sequencer?

dshiell: I guess it must be this way since that won't necessarily hold if the previous state was based off a different L1 view

dshiell: Would it help to update ILightClient to allow the l1header field to be `bytes` and then allow the LC implementation to cast it to the desired struct?

contracts/core/SequencerProofVerifier.sol
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    /**
+     * @dev Verifies an ICS23 proof through the root hash based on the provided proof.
+     * @dev This code was adapted from the ICS23 membership verification found here:
+     * https://github.com/cosmos/ics23/blob/go/v0.10.0/go/ics23.go#L36
+     * @param proof The ICS23 proof to be verified.
+     * @return computed The computed root hash.
+     */
+    function _verify(OpIcs23Proof calldata proof) internal pure returns (bytes32 computed) {
```
nicopernas, line 107: this looks a lot like the one from OptimisticProofVerifier.sol, can we put them in a lib instead of duplicating the code?

nicopernas: same with the one below

contracts/core/SequencerProofVerifier.sol
nicopernas: +1 to not using the full L1 header here. We had that field in the original "merkle proof client" because it was the only way to link the L1 state root to a L1 hash that we could then compare with the chain's view of L1 (via L1Block contract) but in here we trust the L1 hash as long as the signature matches.

RnkSngh: would something like[ this commit](https://github.com/open-ibc/vibc-core-smart-contracts/pull/198/commits/62a1e2645af3d7100ed64edf4ba0decdd31217ed) work?

contracts/core/SequencerProofVerifier.sol
RnkSngh: yes, I need to rebase the branch & update SequencerProofVerifier, but I implemented abstracting this away in [this commit](https://github.com/open-ibc/vibc-core-smart-contracts/pull/198/commits/1ae7826a03ad729d20a063cc59912698c8297648) on the stacked pr branch. 

Note: I chose implementing this functionality in an abstract contract instead of a library since that would cause an extra runtime `DELEGATECALL` gas cost for these methods, since solidity implements libraries as delegate calls. This doesn't happen with abstract contracts since it's compiled into the bytecode of the inheriting contracts

contracts/core/SequencerProofVerifier.sol
RnkSngh: see[ this comment](https://github.com/open-ibc/vibc-core-smart-contracts/pull/199#discussion_r1765117411)

contracts/core/SequencerProofVerifier.sol
RnkSngh: should be added now that we have rebased the generic bytes arg

contracts/core/SequencerProofVerifier.sol
RnkSngh: should be abstracted away into `AppstateVerifier` class now

nicopernas (approved):

RnkSngh:
merging this to unblock @nicopernas

PR #202, opened 2024-09-18
Working version of making acks optional

mvpoyatt:
## Summary by CodeRabbit

- **New Features**
	- Introduced `AckStatus` enumeration for improved acknowledgment handling.
	- Added a new `Pluto` contract to demonstrate IBC functionality.

- **Bug Fixes**
	- Enhanced acknowledgment processing to handle different statuses (`SKIP`, `SUCCESS`, `FAILURE`) more robustly.

- **Tests**
	- Updated tests to validate acknowledgment statuses using the new `AckStatus` enumeration instead of boolean values.
	- Added new test cases for the `Pluto` contract to ensure correct acknowledgment behavior.
	- Expanded test coverage for acknowledgment handling across various contracts.

### DIFF
--- a/contracts/libs/Ibc.sol
+++ b/contracts/libs/Ibc.sol
@@ -65,10 +65,16 @@ struct Height {
     uint64 revision_height;
 }
 
+enum AckStatus {
+    FAILURE,
+    SUCCESS,
+    SKIP
+}
+
 struct AckPacket {
     // success indicates the dApp-level logic. Even when a dApp fails to process a packet per its dApp logic, the
     // delivery of packet and ack packet are still considered successful.
-    bool success;
+    AckStatus status;
     bytes data;
 }
 
@@ -245,8 +251,8 @@ library Ibc {
     function parseAckData(bytes calldata ack) external pure returns (AckPacket memory ackData) {
         // this hex value is '"result"'
         ackData = (keccak256(ack[1:9]) == keccak256(hex"22726573756c7422"))
-            ? AckPacket(true, Base64.decode(string(ack[11:ack.length - 2]))) // result success
-            : AckPacket(false, ack[10:ack.length - 2]); // this is an error
+            ? AckPacket(AckStatus.SUCCESS, Base64.decode(string(ack[11:ack.length - 2]))) // result success
+            : AckPacket(AckStatus.FAILURE, ack[10:ack.length - 2]); // this is an error
     }
 
     function toStr(bytes32 b) public pure returns (string memory outStr) {
--- a/test/Dispatcher/Dispatcher.t.sol
+++ b/test/Dispatcher/Dispatcher.t.sol
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: UNLICENSED
 pragma solidity ^0.8.13;
 
-import {Ibc, ChannelEnd, IbcEndpoint, Height} from "../../contracts/libs/Ibc.sol";
+import {Ibc, ChannelEnd, IbcEndpoint, Height, AckStatus} from "../../contracts/libs/Ibc.sol";
 import {IbcUtils} from "../../contracts/libs/IbcUtils.sol";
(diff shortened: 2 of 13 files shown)


test/Dispatcher/Dispatcher.t.sol
```diff
@@ -350,11 +351,22 @@ contract DispatcherRecvPacketTestSuite is ChannelOpenTestBaseSetup {
...
 
+    // check that for pluto, recvPacket does not trigger an ack
+    function test_pluto_noAck() public {
+        uint64 packetSeq = 1;
+        Pluto pluto = new Pluto(dispatcherProxy);
+        string memory plutoPort = string(abi.encodePacked(portPrefix, getHexBytes(address(pluto))));
+        IbcEndpoint memory plutoDest = IbcEndpoint(plutoPort, channelId);
+        vm.expectEmit(true, true, true, true, address(dispatcherProxy));
+        emit RecvPacket(address(pluto), channelId, packetSeq);
+        dispatcherProxy.recvPacket(IbcPacket(src, plutoDest, packetSeq, payload, ZERO_HEIGHT, maxTimeout), validProof);
```
RnkSngh, line 367: in addition to the asserting that the recvPacket event was emitted, we also need to test that:
- the `WriteAckPacket` was NOT emitted 
- `_ackPacketCommitment` was NOT set to true

for the first case, we have to be a bit clever, use [expectEmit](https://book.getfoundry.sh/forge/cheatcodes?highlight=emit#cheatcodes) and emit a unique event after the call, such that the expect emit will pass iff `writeAckPacket` event was not emitted (I'd suggest reading expect emit docs first to understand how event testing works, it's a tad unintuitive) 

for second case, you can use something like `findAckPacketCommitmentSlot` in `test/upgradeableProxy/UpgradeUtils.t.sol` to calculate and read from a storage slot in dispatcher proxy, and ensure that that is indeed not set to `true`

contracts/libs/Ibc.sol
```diff
@@ -65,10 +65,16 @@ struct Height {
...
+    FAILURE,
+    SUCCESS,
+    SKIP
+}
+
 struct AckPacket {
     // success indicates the dApp-level logic. Even when a dApp fails to process a packet per its dApp logic, the
     // delivery of packet and ack packet are still considered successful.
-    bool success;
+    AckStatus status;
```
RnkSngh, line 77: I'm just now realizing that, because this `AckPacket` type is used in the acknowledgement proof, we can't change this existing type and need to add a new type with the enum. Otherwise we might have to make changes to peptide to support this new enum change. 

So I think we need to have two seperate `AckPackets` types now:
- an `AckPacket` with the enum status which the dapp returns in the onrecv callback
- the original `AckPacket` type with the bool success that is emitted in the dispatcher events (and in turn used in peptide proofs)

contracts/libs/Ibc.sol
```diff
@@ -245,8 +251,8 @@ library Ibc {
     function parseAckData(bytes calldata ack) external pure returns (AckPacket memory ackData) {
         // this hex value is '"result"'
         ackData = (keccak256(ack[1:9]) == keccak256(hex"22726573756c7422"))
-            ? AckPacket(true, Base64.decode(string(ack[11:ack.length - 2]))) // result success
-            : AckPacket(false, ack[10:ack.length - 2]); // this is an error
+            ? AckPacket(AckStatus.SUCCESS, Base64.decode(string(ack[11:ack.length - 2]))) // result success
+            : AckPacket(AckStatus.FAILURE, ack[10:ack.length - 2]); // this is an error
```
RnkSngh, line 255: See the other comment, but we actually can't change the `AckPacket` return type here, since `parseAckData` is used to find the proof key in the `acknowledge` method in dispatcher. We need to have two seperate ackPacket types now

RnkSngh (changes requested):

PR #235, opened 2024-10-23
Raunak/multi solc artifacts

RnkSngh:
Our vibc core artifact files are always in the format of `FactoryName.sol/FactoryName.json`. 

However, some Otimism contract artifacts are in the format of `Factoryname.sol<SolcVersion>.sol/FactoryName.json<SolcVersion>`, since some are compiled using multiple versions of solc. 

This PR adds support for this new format in our deploy script, so that we can deploy 1 contracts from our infra.

### DIFF
--- a/src/deploy.ts
+++ b/src/deploy.ts
@@ -9,7 +9,7 @@ import {
 } from "./utils/io";
 import assert from "assert";
 import {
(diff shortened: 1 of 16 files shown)


dshiell (approved):
lgtm

PR #238, opened 2024-10-25
Raunak/account registry cleanup

RnkSngh:
This PR Separates sendingAccounts into two different registries. On one hand, having the account registry return both single and multisig types is not great for contexts where we only need single sig accounts (e.g. e2e) since we have to typeguard everywhere to properly parse data. 

OTOH, to make infra integration easier, it would be nice to have support for being able to specify multiple accounts types (multisig vs single sig) from a single account spec file. 

This PR solves this by having two exported account registry types - one called `SingleSigAccountRegistry` and one called `SendingAccountRegistry`. We can move our e2e from using the generic account registry to a `SingleSigAccountRegistry`, while having our infra use the `SendingAccountRegistry`

### DIFF
--- a/src/deploy.ts
+++ b/src/deploy.ts
@@ -9,7 +9,7 @@ import {
 } from "./utils/io";
 import assert from "assert";
 import {
-  AccountRegistry,
+  SingleSigAccountRegistry,
   connectProviderAccounts,
   Wallet,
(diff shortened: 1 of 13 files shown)


dshiell (approved):
lgtm

PR #239, opened 2024-10-25
Raunak/multisig from spec

RnkSngh:
Instead of having to pass in all arguments (owners, threshold, signing privatekey , chainId) needed to create a multisg , this PR adds support for specifying an initialized and uninitialized multisig type in accounts spec. This makes the infra integration a lot easier


dshiell (approved):
lgtm

PR #258, opened 2024-11-21
Raunak/proof api clients

RnkSngh:
## Summary by CodeRabbit

- **New Features**
	- Introduced the `CrossL2Prover` contract for validating cross-layer receipts and events.
	- Added functions for receipt and event validation, state retrieval, and client updates.
	- New interfaces and bindings for `ICrossL2Prover` and `CrossL2Prover` added for enhanced contract interaction.

- **Documentation**
	- Enhanced documentation for multiple functions across contracts to improve usability and clarity.

- **Version Update**
	- Updated version number in `package.json` from `4.0.10` to `4.0.11`.


contracts/interfaces/ILightClient.sol
```diff
@@ -21,18 +21,14 @@ import {Ics23Proof} from "./IProofVerifier.sol";
 enum LightClientType {
     SimTestLightClient, // Note: not deployed on any mainnets
     OptimisticLightClient, // Our native opstack light client
-    SequencerLightClient // Our native sequencer light client
+    SequencerLightClient, // Our native sequencer light client, which does not check l1 origin check to cut down on
```
dshiell, line 24: Can we/should we combine these and use a bool at deploy time to configure L1 checks?

contracts/libs/PolymerProofs.sol
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+pragma solidity ^0.8.0;
+
+/**
+ * A library for helpers for proving peptide state
+ */
+library PolymerProofs {
+    function receiptProofKey(bytes32 clientId, uint64 height) internal pure returns (bytes memory proofKey) {
+        // {chainId}/trustedReceiptRoot/{height}
```
dshiell, line 25: this might be `polyibc/client/<clientid>/trustedReceiptRoot/height` ?

contracts/core/proofAPI/CrossL2EventProver.sol
```diff
@@ -0,0 +1,109 @@
...
+    error CannotUpdateClientWithDifferentAppHash();
+
+    constructor(ISignatureVerifier verifier_) {
+        verifier = verifier_;
+    }
+
+    /**
+     * @inheritdoc ICrossEventProver
+     */
+    function validateReceipt(
```
dshiell, line 48: When is this called?

contracts/core/CrossL2Inbox.sol
```diff
@@ -0,0 +1,78 @@
...
+    ///         and emits an ExecutingMessage event. This function is useful
+    ///         for applications that understand the schema of the _message payload and want to
+    ///         process it in a custom way.
+    /// @param _root receipt root.
+    /// @param _key receipt index.
+    /// @param _value raw receipt.
+    /// @param _receiptProof receipt proof.
+    /// @param _logHash Hash of the raw log.
+    /// @param _logIdx raw log index.
+    function validateMessageWithProof(
```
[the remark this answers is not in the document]
RnkSngh: this was Bo's old poc code, I think we can get rid of the `CrossL2Inbox.sol` altogether

contracts/interfaces/ILightClient.sol
dshiell: SequencerLightClient and ReOrgResistantSequencerLightClient

contracts/core/SequencerSoloClient.sol
```diff
@@ -75,7 +77,9 @@ contract SequencerSoloClient is ILightClient {
     }
 
     /**
-     * @inheritdoc ILightClient
+     * @dev Checks if the current trusted optimistic consensus state
+     * can be used to perform the membership test and if so, verifies the proof
+     * @dev reverts if the proof is not valid (i.e. if the key is not included in the proof)
      */
     function verifyMembership(Ics23Proof calldata proof, bytes calldata key, bytes calldata expectedValue)
```
i-norden, line 84: Did we find out what the reason for using AppHash at n-1 for proof at n is?

```
// a proof generated at height H can only be verified against state root (app hash) from block H - 1.
// this means the relayer must have updated the contract with the app hash from the previous block and
// that is why we use proof.height - 1 here.
```

contracts/core/CrossL2Inbox.sol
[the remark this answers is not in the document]
i-norden: Agree!

contracts/core/SequencerSoloClient.sol
RnkSngh: we didn't go super down into the weeds, but I think it was something to do with how ics23 proofs work, and that the relayer was assuming to use `height - 1` so it was still consistent with what the relayer expects

contracts/core/proofAPI/CrossL2EventProver.sol
```diff
@@ -0,0 +1,109 @@
...
+            bytes32 peptideClientId
+        ) = abi.decode(eventProof, (Heights, bytes, Ics23Proof, bytes[], bytes32));
+        // Before we can trust the receipt root, we first need to verify that the receipt root is indeed stored
+        // on peptide at the given clientID and height.
+
+        // VerifyMembership verifies the receipt root  through an ics23 proof of peptide state that attests that the
+        // given sourceL2Height has the receipt root at the peptide height
+        this.verifyMembership(
+            bytes32(_getPeptideAppHash(peptideAppProof.height)),
+            PolymerProofs.receiptProofKey(peptideClientId, heights.sourceL2Height),
```
dshiell, line 63: verifier contract needs to check client id.

contracts/core/SequencerSoloClient.sol
dshiell: I believe this is because you can only prove things had already happened and are in the app state which isn't true until you've completed the block and are at the next height. i.e A proof at height, H, uses the merkle tree of app state built a height, H - 1

contracts/core/proofAPI/CrossL2EventProver.sol
```diff
@@ -0,0 +1,145 @@
...
+
+    constructor(ISignatureVerifier verifier_, string memory counterpartyClientID_) {
+        verifier = verifier_;
+        counterpartyClientID = counterpartyClientID_;
+    }
+
+    /**
+     * @inheritdoc ICrossEventProver
+     */
+    function validateReceipt(bytes calldata receiptIndex, bytes calldata receiptRLPEncodedBytes, bytes calldata proof)
```
dshiell, line 50: Should these be "memory" if often call from another contract? Same would apply to validateEvent below.

contracts/core/proofAPI/CrossL2EventProver.sol
```diff
@@ -0,0 +1,145 @@
...
+            uint64 sourceL2Height
+        ) = abi.decode(proof, (Ics23Proof, bytes[], bytes32, bytes32, uint64));
+        // Before we can trust the receipt root, we first need to verify that the receipt root is indeed stored
+        // on peptide at the given clientID and height.
+
+        // VerifyMembership verifies the receipt root  through an ics23 proof of peptide state that attests that the
+        // given sourceL2Height has the receipt root at the peptide height
+        this.verifyMembership(
+            bytes32(_getPeptideAppHash(peptideAppProof.height)),
+            PolymerProofs.receiptProofKey(peptideClientId, sourceL2Height),
```
dshiell, line 69: please add a note that we should be verifying peptideClientId matches the contract clientid

contracts/core/proofAPI/CrossL2EventProver.sol
dshiell: Side note, having the proof be plain bytes here is nice if we ever used a zkp version of the CrossL2EventProver contract in the future.

contracts/interfaces/ICrossEventProver.sol
```diff
@@ -0,0 +1,96 @@
...
+    // -  an Ics23 proof object that proves a receipt root is stored in an ival tree at a given height
+    // - a bytes[] object that contains the MMPT proof needed to prove the receipt is included in the receipt root
+    // - a bytes32 peptideClientId
+    struct EventProof {
+        Ics23Proof iavlProof; // Proof that the receipt root exists within an iavl tree at a given apphash stored in
+            // this contract
+        bytes[] receiptMMPTProof; // A modified Merkle Patricia tree proof that proves the raw RLP receipt bytes are
+            // included in the receipt root
+        bytes32 receiptRoot; // The receipt root that is is proven against in the MMPT proof
+        uint64 sourceL2Height; // The source L2 block the receipt or event we are trying to prove was emitted in. The
```
dshiell, line 38: nit: should we call it "eventHeight" ?

dshiell (approved):
ship it!

contracts/interfaces/ICrossEventProver.sol
RnkSngh: sure

contracts/core/proofAPI/CrossL2EventProver.sol
RnkSngh: added!

contracts/core/proofAPI/CrossL2EventProver.sol
RnkSngh: peptide client id should be the hardcoded one per contract

PR #270, opened 2024-12-02
generate go bindings in one package per contract/library

RnkSngh:
PR to make each contract _wihtin each sol file_ have it's own module. 

Previously, if we ran the `make bindings-gen-go` command, it would produce go files which would not build for the mars contract, since it would declare all contracts declared in the `contracts/examples/Mars.sol` (Mars, PanicingMars, etc) as part of one go module. This would produce errors if you tried to build the go bindings using `go build ./...` because it would result in doubly-declaring certain structs (e.g. AckPacket for Mars contracts). To allow our go bindings to be robust to having multiple contracts to be declared in the same sol files, we can declare each contract as part of it's own package. 

This should only impact users of the reverting/panicking mars (of which there are none), and  `IbcDispatcher` go bindings

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,10 +62,11 @@ bindings-gen-go: build-contracts
 			continue; \
 		fi; \
 		type=$$(basename $$abi_file .abi.json); \
-		pkg=$$(basename $$abi_base .sol | tr "[:upper:]" "[:lower:]"); \
+		pkg=$$(basename $$type .sol | tr "[:upper:]" "[:lower:]"); \
 		mkdir -p ./bindings/go/$$pkg; \
 		abigen --abi $$abi_file --pkg $$pkg --type $$type --out ./bindings/go/$$pkg/$$type.go; \
 	done; \
+	go build ./... || exit 1; \
 	echo "Done."
 
 bindings-gen-ts: build-contracts
--- a/bindings/go/iappstateverifier/IAppStateVerifier.go
+++ b/bindings/go/iappstateverifier/IAppStateVerifier.go
@@ -1,7 +1,7 @@
 // Code generated - DO NOT EDIT.
 // This file is a generated binding and any manual changes will be lost.
 
-package iproofverifier
+package iappstateverifier
 
 import (
 	"errors"
--- a/bindings/go/ibceventsemitter/IbcEventsEmitter.go
+++ b/bindings/go/ibceventsemitter/IbcEventsEmitter.go
@@ -1,7 +1,7 @@
 // Code generated - DO NOT EDIT.
 // This file is a generated binding and any manual changes will be lost.
 
-package ibcdispatcher
+package ibceventsemitter
 
 import (
 	"errors"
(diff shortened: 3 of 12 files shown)


Makefile
```diff
@@ -62,10 +62,11 @@ bindings-gen-go: build-contracts
 			continue; \
 		fi; \
 		type=$$(basename $$abi_file .abi.json); \
-		pkg=$$(basename $$abi_base .sol | tr "[:upper:]" "[:lower:]"); \
+		pkg=$$(basename $$type .sol | tr "[:upper:]" "[:lower:]"); \
 		mkdir -p ./bindings/go/$$pkg; \
 		abigen --abi $$abi_file --pkg $$pkg --type $$type --out ./bindings/go/$$pkg/$$type.go; \
 	done; \
+	go build ./... || exit 1; \
```
nicopernas, line 69: nit: you need the `|| exit 1`. If `go build` fails, the make target will fail too.

nicopernas (approved):

Makefile
nicopernas: nit2: maybe add a message saying "running check sanity on bindings" or similar?

RnkSngh: (assuming you meant I _don't_ need the `exit 1` ) 

I think we actually do need this `exit 1` to cause the root workflow to fail ; not sure if this is default behavior or maybe it's an os thing but the make command doesn't fail of the go build fails without the exit 1 - I tried this myself through replacing the `go build ./...` with `go build ./.` to cause it to fail due to not finding any files to build. Running the failing command without the exit command still causes the overall make command to pass. 

Similar to the behavior we're seeing with abigen

RnkSngh: re second nit though I added messaging as you requested, thanks for the suggestion

nicopernas: Sorry, I did mean "don't need" LOL

`Make` should propagate the errors thrown by shell commands executed by targets. Note that each line within a target is handled by a subshell. In your case you only have one command (single subshell) since you are concatenating commands like so `cmd0; cmd1; cmd2; ...` in which case you do have to exit

Here's a clear example 
```
[nicolas@macbook foo]# cat Makefile
fail:
        @go build ./.
shell:
        @echo hello; \
        go build ./.; \
        echo bye;
exit:
        @echo hello; \
        go build ./. || exit 1; \
        echo bye;

[nicolas@macbook foo]# make fail ; echo ">> $?"
package .: no Go files in /private/tmp/foo
make: *** [fail] Error 1
>> 2

[nicolas@macbook foo]# make shell ; echo ">> $?"
hello
package .: no Go files in /private/tmp/foo
bye
>> 0

[nicolas@macbook foo]# make exit ; echo ">> $?"
hello
package .: no Go files in /private/tmp/foo
make: *** [exit] Error 1
>> 2
```

Things like for loops need to be within a single subshell (so you need the `; \`) but other things could be in separate ones. Not a big deal though - it's ok as is.

PR #278, opened 2024-12-06
add example proof api contract

RnkSngh:
## Summary by CodeRabbit

- **New Features**
	- Introduced the `Venus` smart contract for validating events across blockchains, including receipt and event handling.
	- Added events for successful receipt and event processing.
	- Implemented error handling for receipt and event proof validation.
	- Created TypeScript and Go bindings for the `Venus` contract, enhancing integration capabilities.

- **Chores**
	- Updated the `Makefile` to include the new `Venus` contract in the `CONTRACT_NAMES` variable.
	- Updated `package.json` to reflect a new version number.


contracts/examples/Venus.sol
```diff
@@ -0,0 +1,74 @@
...
+        external
+    {
+        if (!prover.validateReceipt(receiptIndex, receiptRLPEncodedBytes, proof)) {
+            revert invalidReceiptProof();
+        }
+
+        emit SuccessfulReceipt(receiptIndex, receiptRLPEncodedBytes);
+    }
+
+    function checkEventProof(
```
alfredo-stonk, line 55: nit: rename these methods to `receiveReceipt` and `receiveEvent` so they convey meanings from dApps' perspective? some comments would be nice to, eg. solvers would call these methods, add dapp logic here, etc

contracts/examples/Venus.sol
```diff
@@ -0,0 +1,74 @@
...
+ * @title Mars
+ * @notice Mars is a simple IBC receiver contract that receives packets and sends acks.
+ * @dev This contract is used for only testing IBC functionality and as an example for dapp developers on how to
+ * integrate with the vibc protocol.
+ */
+contract Venus {
+    ICrossL2Prover public immutable prover ;
+
+    event SuccessfulReceipt(bytes receiptIndex, bytes receiptRLP);
+    event SuccessfulEvent(uint eventIndex, bytes eventBytes);
```
alfredo-stonk, line 32: add msg.sender or tx.origin as event fields too? receiptRLP or eventBytes may not be necessary as they're identical to calldata

alfredo-stonk (approved):

contracts/examples/Venus.sol
RnkSngh: valid. I remove rlp byes and substitute for sender. Maybe receipt root is more useful here as well

RnkSngh: actually, I think it might be slightly complicated to find the original sender of the log, since we have to json decode it?

contracts/examples/Venus.sol
RnkSngh: sure! changed

contracts/examples/Venus.sol
RnkSngh: ah you meant msg.sender of the dest chain, yes i can add

PR #281, opened 2024-12-10
Feat : Create a basic contract for proof api health check

OBlackmon3:
## What?
Create a smart contract to generate events for Proof API
## Why?
We need a consistent source of events to prove on other chains
## How?
The contract has a method that accepts transactions
## Testing?
N/A

## Summary by CodeRabbit

- **New Features**
	- Introduced a smart contract for managing Ether transactions and logging events.
	- Added functionality for receiving Ether and sending it to specified recipients.
	- Implemented event logging for transaction details, including health checks and transfers.
	- Provided a method to retrieve the current Ether balance of the contract.

### DIFF
--- a/contracts/utils/EventCreation.sol
+++ b/contracts/utils/EventCreation.sol
@@ -1,0 +1,37 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+contract SimpleTransaction {
+    address public owner;
+
+    // Event to log the transaction
+    event EtherTransferred(address indexed to, uint256 amount);
+
+    // Constructor to set the contract owner
+    constructor() {
+        owner = msg.sender;
+    }
+
+    // Function to receive Ether
+    receive() external payable {}
+
+    // Function to send Ether from the contract to a specified address
+    function sendEther(address payable recipient, uint256 amount) public {
+        // Ensure only the owner can send Ether
+        require(msg.sender == owner, "Only the owner can call this function");
+
+        // Check that the contract has enough balance
+        require(address(this).balance >= amount, "Insufficient contract balance");
+
+        // Transfer the Ether
+        recipient.transfer(amount);
+
+        // Emit an event for logging
+        emit EtherTransferred(recipient, amount);
+    }
+
+    // Function to get the contract's balance
+    function getBalance() public view returns (uint256) {
+        return address(this).balance;
+    }
+}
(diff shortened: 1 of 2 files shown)


contracts/utils/EventCreation.sol
```diff
@@ -0,0 +1,37 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+contract SimpleTransaction {
```
RnkSngh, line 4: nit: like i don't have strong opinions on this or anything but we did kinda have a planet naming thing going on for example contracts

but I can't emphasize enough. I don't have any strong opinions on this. 

But we did have a planet thing going on

contracts/utils/EventCreation.sol
```diff
@@ -0,0 +1,37 @@
...
+    address public owner;
```
RnkSngh, line 5: consider inheriting from [openzeppelin's Ownable](https://docs.openzeppelin.com/contracts/2.x/access-control#ownership-and-ownable) & use `onlyOwner` modifier for sending

RnkSngh (approved):

RnkSngh:
might be nice if we had some sort of dynamic event emitting. Wondereing if we can do `emit 0x...` with arbitrary bytes or if compiler won't like that

contracts/utils/ProofAPIEventGenerator.sol
RnkSngh, line 1: hmm let's put it in the same directory as the other example contracts like mars & venus 

`contracts/examples/`

contracts/utils/ProofAPIEventGenerator.sol
```diff
@@ -0,0 +1,40 @@
...
+    // Event to log the transaction
+    event ProofAPIEventHealthCheck(address indexed to, uint256 amount, uint256 srcChainID, uint256 destChainID);
+
+    // Constructor to set the contract owner
+    constructor() {
+        owner = msg.sender;
+    }
+
+    // Function to receive Ether
+    receive(uint256 amount uint256 srcChainID, uint256 destChainID) external payable {
```
RnkSngh, line 16: make sure code compiles using `forge build`
```suggestion
    receive(uint256 amount, uint256 srcChainID, uint256 destChainID) external payable {
```

RnkSngh (approved):

contracts/utils/EventCreation.sol
OBlackmon3: Will do!

contracts/utils/EventCreation.sol
OBlackmon3: Thank you. Will rename to mercury

OBlackmon3:
No longer needed for now

PR #282, opened 2024-12-10
Feat : Add a health check event and function to Venus

OBlackmon3:
## What?
Add a health check event and function to Venus
## Why?
We are using this as a source of events to prove on other chains
## How?
Created a new function and event to emit. The event is uniquely identified by the hash of the parameters
## Testing?
N/A

## Summary by CodeRabbit

- **New Features**
	- Introduced a new event, `SendHealthCheckEvent`, to enhance contract interactions.
	- Added a public function, `emitEvent`, allowing users to emit health check events with unique identifiers based on chain IDs and transaction details.


RnkSngh (approved):
LGTM, but seems like our CI is failling. 

(we should prob have these as git hooks but in the mean time you'll have to do the following manually)
- run `forge build` to make sure contracts can comile
- run `forge fmt && npm run lint` to check linter
- build bindings  using `npm run build`

contracts/examples/Venus.sol
```diff
@@ -30,6 +30,7 @@ contract Venus {
 
     event SuccessfulReceipt(bytes receiptIndex, bytes receiptRLP);
     event SuccessfulEvent(uint256 eventIndex, address sender);
+    event SendHealthCheckEvent(bytes32 id, uint256 sourceChainID, uint256 destChainID);
```
dshiell, line 33: Lets call it SendHealthCheck or SendMessage. 'Event' is redundant here.

RnkSngh (approved):
Yay looks like you got bindings to pass! 

I bless it. 

let's bump "version" property in package json to 4.0.16 and then mergeee

OBlackmon3:
No longer needed for now

PR #287, opened 2024-12-16
update interface

RnkSngh:
## Release Notes

- **New Features**
  - Introduced `clientType` method to retrieve client type.
  - Added new methods for log validation and receipt validation with simplified parameters.
  - Enhanced `Venus` contract with new state variables and event handling for cross-chain interactions.
  - New error types added for improved error handling in the `Venus` contract.

- **Bug Fixes**
  - Updated method signatures for receipt and event validation to streamline functionality.

- **Documentation**
  - Updated ABI definitions to reflect new methods and parameters.

- **Chores**
  - Incremented version number from `4.0.15` to `4.0.17`.


contracts/libs/Ibc.sol
```diff
@@ -286,7 +286,11 @@ library Ibc {
...
     }
 
-    function receiptRootKey(string memory clientId, uint256 height) internal pure returns (bytes memory proofKey) {
-        proofKey = abi.encodePacked("client/", clientId, "/receiptRoot/", toStr(height));
+    function receiptRootKey(bytes32 chainId, bytes32 clientType, uint256 height)
+        internal
+        pure
+        returns (bytes memory proofKey)
+    {
+        proofKey = abi.encodePacked("chain/", chainId, "/client/", clientType, "/receiptRoot/", toStr(height));
```
dshiell, line 294: need to verify this so we're in alignment across peptide/relayer/contracts.

see Jack's PR: https://github.com/polymerdao/polymerase/pull/2036/files?diff=split&w=1#diff-bf10a17910f29c052bfd060654113f39c7e9e6f1db4adcbc0538309413cd9501R77

dshiell (commented):
Looks good! only things are to move from 'draft' status and then check the receiptRoot key with @jlafiandra6

contracts/libs/Ibc.sol
nicopernas: Let's use `chain/<chainId>/client/<clientType>/receiptRoot/<height>`

nicopernas: also, the `chainID` and the `clientType` should be strings (as `height` is)

contracts/core/proofAPI/CrossL2Prover.sol
```diff
@@ -31,29 +31,31 @@ contract CrossL2Prover is AppStateVerifier, ICrossL2Prover {
...
-    {
-        (Ics23Proof memory peptideAppProof, bytes[] memory receiptMMPTProof, bytes32 receiptRoot, uint64 eventHeight) =
-            abi.decode(proof, (Ics23Proof, bytes[], bytes32, uint64));
+    function validateReceipt(bytes calldata proof) public view returns (bytes32 srcChainID, bytes memory receiptRLP) {
+        (
+            Ics23Proof memory peptideAppProof,
+            bytes[] memory receiptMMPTProof,
+            bytes32 receiptRoot,
+            uint64 eventHeight,
+            bytes32 srcChainId,
```
nicopernas, line 56: this needs to match what peptide is doing: https://github.com/polymerdao/relayer/pull/17/files#diff-272397fd750808e0c4740b9200992b4bc15a0cacfd75cb4360bef55ef743b43eR203
ie it should be a string encoded number

contracts/libs/Ibc.sol
RnkSngh: I was assuming chainId was already going to be encoded as bytes, but we can leave that in for a future optimizaiton

RnkSngh: will change it to be back into a string

contracts/core/proofAPI/CrossL2Prover.sol
RnkSngh: done!

dshiell (approved):
lets do it

